PR #13, opened 2021-05-26
Add save temp images and final stack

fdrgsp:


micromanager_gui/main_window.py
```diff
@@ -148,92 +156,105 @@ def add_frame_explorer(self, name, array):
...
-            #set which dimension to display in napari viewer -> viewer.dims.set_point(axis, index)
-            self.viewer.dims.set_point(3, event_index_c)
-            self.viewer.dims.set_point(2, event_index_z)
-            self.viewer.dims.set_point(1, event_index_p)
-            self.viewer.dims.set_point(0, event_index_t)   
+            #save layer when acquisition is finished
+            if self.mda.save_groupBox.isChecked():
 
-        except KeyError:
+                if tuple(i+1 for i in im_idx) == seq.shape:
```
tlambert03, line 206: if you want this all to happen when the sequence is finished, this shouldn't go in `_on_mda_frame`... instead you should connect to the [onMDAFinished](https://github.com/tlambert03/pymmcore-remote/blob/044ff0730949725bac97a167fa63b2f0ed71c9c5/pymmcore_remote/_mmcore_plus.py#L114) signal.  (If you need that signal to provide the sequence object, you could submit a PR for that

micromanager_gui/main_window.py
```diff
@@ -148,92 +156,105 @@ def add_frame_explorer(self, name, array):
...
+                    save_path = Path(self.mda.dir_lineEdit.text()) / name
+            
+                    self.viewer.layers[name].save(str(save_path))
+
+
+        except StopIteration:
+
+            if self.mda.save_groupBox.isChecked():
+                file_name = self.mda.fname_lineEdit.text()
+                layer_name = f"{file_name}_{datetime.now().strftime('%H:%M:%S')}"
```
tlambert03, line 219: let's not include the entire filename in the layer name.  If anything, use `os.path.basename`

micromanager_gui/main_window.py
```diff
@@ -148,92 +156,105 @@ def add_frame_explorer(self, name, array):
...
+
+            _image = image[(np.newaxis,) * len(seq.shape)]
+            layer = self.viewer.add_image(_image, name=layer_name)
+            labels = [i for i in seq.axis_order if i in event.index] + ["y", "x"]
+
+            self.viewer.dims.axis_labels = labels
+            layer.metadata["useq_sequence"] = seq
+            layer.metadata["uid"] = seq.uid
+
+            layer = self.viewer.layers[layer_name]
```
tlambert03, line 231: you already have a handle on the layer from 224 ... why get it again?

micromanager_gui/main_window.py
```diff
@@ -148,92 +156,105 @@ def add_frame_explorer(self, name, array):
...
+
+            # get the index of the incoming image
+            im_idx = tuple(event.index[k] for k in seq.axis_order if k in event.index)
+            
+            #save first image in the temp folder
```
tlambert03, line 236: you're duplicating code from above... which is always a good sign that you should extract something into a helper function

micromanager_gui/main_window.py
```diff
@@ -148,92 +156,105 @@ def add_frame_explorer(self, name, array):
...
+    def _on_mda_frame(self, image: np.ndarray, event: useq.MDAEvent):
 
-        sequence = event.sequence
+        seq = event.sequence
 
-        event.index.setdefault('t', 0)
-        event.index.setdefault('z', 0)
+        #create temp folder if it doesnt exist using uid
+        tmp_uid_name = Path('tmp_folder') / str(seq.uid)
+        temp_folder = Path(__file__).parent.parent / tmp_uid_name
```
tlambert03, line 166: don't create temp folders yourself... use [`tempfile.TemporaryDirectory`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory)

micromanager_gui/main_window.py
```diff
@@ -148,92 +156,105 @@ def add_frame_explorer(self, name, array):
...
+            # get the index of the incoming image
+            im_idx = tuple(event.index[k] for k in seq.axis_order if k in event.index)
+            
+            #save first image in the temp folder
+            image_name = f'{im_idx}.tif'
+            savefile = temp_folder / image_name
+
+            tifffile.tifffile.imsave(str(savefile), image)
+
+            if tuple(i+1 for i in im_idx) == seq.shape:
```
tlambert03, line 242: same comment as above... use the mdaFinished signal instead

tlambert03:
#14 is merged... so you can checkout `main`, pull upstream, and then merge it into this branch

micromanager_gui/main_window.py
fdrgsp: but if we need to access the temp folder isn't it easier to have it in the plugin path?

tlambert03: not sure I'm following the reasoning... You can still access a TemporaryDirectory using its `.name` attribute (read the docs I linked above)

fdrgsp: The problem that I have is that the name of the temp folder keeps changing every time _on_mda_frame() is called.

Even if I specify prefix and suffix :
`temp_dir = tempfile.TemporaryDirectory(None, str(seq.uid))`
`print(temp_dir.name)` -> ../44e3f048-bbeb-4bc9-88b8-a1cdc6903dc7**8b44f8jh** 
where 8b44f8jh is added and change every time.

is doing something like that that wrong:

```
try:
    temp_folder = Path(tempfile.gettempdir()), ('uid_' + str(seq.uid))
    Path(temp_folder).mkdir(parents=True, exist_ok=False)
except FileExistsError:
    pass
```

micromanager_gui/main_window.py
fdrgsp: I managed to do it, It's super cool!!! :)

fdrgsp: Do you think is worth it to delete the temp folder also  onMDACanceled?

tlambert03: if you look through the [events emitted by `run_mda`](https://github.com/tlambert03/pymmcore-remote/blob/044ff0730949725bac97a167fa63b2f0ed71c9c5/pymmcore_remote/_mmcore_plus.py#L72) you'll notice that if a sequence is canceled, it just breaks the for loop and goes right to the `self.emit_signal("onMDAFinished")` ... so as long as you hook up onMDAFinished to tempDir deletion, you'll be good

micromanager_gui/main_window.py
tlambert03: just like temp_dir deletion shouldn't be in `_on_mda_frame`, nor should tempdir creation... you should create the tempdir _once_ ... on the `onMDAStarted` started signal.  (just think of `_on_mda_frame` as "stuff that needs to be done on _every_ frame... if something you're trying to do doesn't need to happen on every frame, it doesn't belong in `_on_mda_frame`).  create new methods if you need to

fdrgsp: Thanks Talley, this helped me a lot! I moved everything out of _on_mda_frame

micromanager_gui/main_window.py
```diff
@@ -148,92 +159,121 @@ def add_frame_explorer(self, name, array):
         except KeyError:
             self.viewer.add_image(array, name=layer_name)
 
-    # TO DO: add the file name form the save box
-    def _on_mda_frame(self, image, event):
 
-        sequence = event.sequence
+    # create temp folder when mda starts
+    def _on_mda_started_temp_folder(self, sequence: useq.MDASequence):
```
tlambert03, line 164:
```suggestion
    def _on_mda_started(self, sequence: useq.MDASequence):
        """"create temp folder when mda starts."""
```

for a callback like this, name it based on what it responds to, not what it does (you may want to add things other than creating a temp folder later).

also, use docstrings for comments that aren't things like "TODO:"...

micromanager_gui/main_window.py
```diff
@@ -148,92 +159,121 @@ def add_frame_explorer(self, name, array):
...
 
-        event.index.setdefault('t', 0)
-        event.index.setdefault('z', 0)
+        self.temp_folder = tempfile.TemporaryDirectory(None, str(sequence.uid))
+        return self.temp_folder.name
```
tlambert03, line 167: probably don't need to return anything here?

micromanager_gui/main_window.py
```diff
@@ -148,92 +159,121 @@ def add_frame_explorer(self, name, array):
...
 
-        event_index_t = event.index["t"]
-        event_index_p = event.index["p"]
-        event_index_z = event.index["z"]
-        event_index_c = event.index["c"]
-        
-        file_name = 'mda'
+    # delete temp folder and files when mda is done and save layer
+    # and add increment to save name
+    def _on_mda_finished_save_layer(self, sequence: useq.MDASequence):
```
tlambert03, line 183:
```suggestion
    def _on_mda_finished(self, sequence: useq.MDASequence):
        """Delete temp folder and files when mda is done and save layer.
        
        and add increment to save name
        """
```

micromanager_gui/main_window.py
```diff
@@ -148,92 +159,121 @@ def add_frame_explorer(self, name, array):
...
+            
+        if self.mda.save_groupBox.isChecked():
```
tlambert03, line 185: one problem here is that someone could have unchecked it in the meantime (and they wouldn't expect that it would "undo" saving the currently running experiment) ... so we might need to either store this on the sequence, or check for it in `_on_mda_started` and cache it for later

micromanager_gui/main_window.py
```diff
@@ -148,92 +159,121 @@ def add_frame_explorer(self, name, array):
...
 
-        layer_name = (
-            f'{file_name}_[{sequence.axis_order}]_p{p_stack_length}_'
-            f't{t_stack_length}_z{z_stack_length}_c{c_stack_length}'
-        )
+            active_layer = self.viewer.layers.selection.active.name
```
tlambert03, line 187: this makes the assumption that the user hasn't changed the layer in the meantime

micromanager_gui/main_window.py
```diff
@@ -148,92 +159,121 @@ def add_frame_explorer(self, name, array):
...
+            fname =  self.mda.fname_lineEdit.text()
 
-            if sequence.axis_order == 'tpzc' or sequence.axis_order == 'ptzc':            
-                #channels
-                if event_index_c > 0 and event_index_z == 0 and event_index_p == 0 and event_index_t == 0:
-                    empty_im = np.empty(((1,)*len(sequence.axis_order) + image.shape), dtype=np.uint16) 
-                    layer.data = np.concatenate((layer.data,empty_im), axis=-3)
+            save_path = Path(self.mda.dir_lineEdit.text())
+
+            list_dir = [str(i).split('/')[-1] for i in save_path.iterdir() \
```
tlambert03, line 193: I recommend looking through the [pathlib](https://docs.python.org/3/library/pathlib.html) docs... They're very useulf.  Here it looks like you're looking for [glob](https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob): `save_path.glob('*.tif')` ... and possibly also [Path.name](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.name): 
```
[p.name for p in save_path.glob("*.tif")]
```

however, i think the whole `get_filename` thing probably should be reworked

micromanager_gui/main_window.py
```diff
@@ -148,92 +159,121 @@ def add_frame_explorer(self, name, array):
...
+                if (str(i).split('/')[-1]).endswith(('.tif', '.tiff'))]
+
+            try:
```
tlambert03, line 196: can you put some comments in here to make it easier for me to quickly determine what you're trying to accomplish?  (see comments in `_on_mda_frame` for an example of what I mean)

tlambert03 (commented):
coming along!  left a few more comments... can you also merge master so that we no longer see the merged code in this diff here?

micromanager_gui/main_window.py
fdrgsp: can we just deactivate the possibility of checking it while the mda is running?

tlambert03: is that better?  or just easier :)

tlambert03: more to the point... it feels like a "patch".  It's definitely reasonable to disable some things while mda is running.  But I wouldn't _just_ disable the `save_groupBox` ... I would think more broadly about what should and shouldn't be interactive while mda is running and enable/disable _all_ of them.  for instance, perhaps the whole mda widget (besides the run/pause/cancel) buttons should be disabled.

micromanager_gui/main_window.py
fdrgsp: I guess this might work:
```
layer_uid = [l.name for l in self.viewer.layers if l.metadata['uid'] == sequence.uid]
active_layer = self.viewer.layers[str(layer_uid[0])]
```

tlambert03: you're already looping through `viewer.layers` ... why not just grab the layer directly rather than just taking the name so that you then index into viewer.layers again...:

```
for layer in self.viewer.layers:
    if layer.metadata.get('uid') == sequence.uid:
        sequence_layer = layer
        break
else:
    raise IndexError("could not find layer corresponding to sequence")
```

note also, it's very important to use `if l.metadata.get('uid')` instead of `if l.metadata['uid']`, since the former will raise an exception if it encounters a layer that doesn't have the `uid` key in the metadata dict (which will basically be every other layer)

tlambert03: alternate syntax:
```python
try:
    active_layer = next(l for l in self.viewer.layers if l.metadata.get('uid') == sequence.uid)
except StopIteration:
    raise IndexError("could not find layer corresponding to sequence")
```

micromanager_gui/main_window.py
fdrgsp: This is basically what I wanted to achieve, I've added few comments.
I'm sure there is a more correct way to do it, and I actually wanted to discuss with you on how you would approach this task :). It does what I want but I think I probably have to start looking at stuff in a different way....

```
"""
               Ckeck the input filename and modify it to add _nnn in the end.
               The get_filename(fname,list_dir) function check if the name 
               is already present in the save_path and incremant the _nnn accordingly.

               filename examples: user input -> output:
                - mda       -> mda_000
                - mda_3     -> mda_003
                - mda_0001  -> mda_001
                - mda1      -> mda1_000
            """
            try:
                n = fname.split('_')[-1]
                
                if len(n) == 3 and int(n)>= 0:
                    fname = self.get_filename(fname,list_dir)
                
                elif len(n) != 3 and len(n) <=4 and int(n) >= 0:
                    s = ''
                    for i in fname.split('_')[:-1]:
                        s = s + i + '_'
                        print(s)
                    print('s', s)
                    fname = s + '{0:03}'.format(int(n))
                    fname = self.get_filename(fname,list_dir)
                
                else:
                    fname = fname + '_000'
                    fname = self.get_filename(fname,list_dir)

            except ValueError:
                fname = fname + '_000'
                fname = self.get_filename(fname,list_dir)
```

This function instead just checks if the filename is already present in the save_path folder and change the _nnn in the end of the filename accordingly:

```
def get_filename(self, fname, list_dir):
        """
            check if the filename_nnn used to save the layer exists
            and increment _nnn accordingly.
        """
        val = int(fname.split('_')[-1])
        while True:
            new_val = '{0:03}'.format(val)
            fname = fname[:-3] + new_val
            if ((fname + '.tif') or (fname + '.tiff')) not in set(list_dir):
                break
            else:
                val += 1
        return fname
```

fdrgsp:
@tlambert03 ready for review

micromanager_gui/main_window.py
tlambert03: if you have addressed these comments are confident with the fix... click the "resolve conversation" button at the bottom of the comment so I know I don't have check to see if it's still there.  If you feel like i might want to remember what the issue was, you can them them, but perhaps add a comment like "fixed here, etc..."   Otherwise, the reviewer needs to do more work to get themselves back in the mindset of what you're working on

micromanager_gui/main_window.py
```diff
@@ -176,17 +306,32 @@ def _on_mda_frame(self, image: np.ndarray, event: useq.MDAEvent):
             for a, v in enumerate(im_idx):
                 self.viewer.dims.set_point(a, v)
 
+            #save each image in the temp folder, 
+            image_name = f'{im_idx}.tif'
+            savefile = Path(self.temp_folder.name) / image_name
```
tlambert03, line 311: test is [failing](https://github.com/tlambert03/napari-micromanager/pull/13/checks?check_run_id=2786248572#step:8:132) with an attribute error because `self.temp_folder` will _only_ exist if another method `_on_mda_started` has been called first ...  (when I wrote the [test](https://github.com/tlambert03/napari-micromanager/blob/c8e1fbba4cc63dee5ac976608e764816fcbc34f9/tests/test_main_window.py#L43), `_on_mda_started` didn't exist)

In general it's a good idea to try to make sure that class methods don't need to be called in "just the right order".  either put a `self.temp_folder = None` thing in the `__init__` function, or check here for `if hasattr(self, 'temp_folder')` ...

micromanager_gui/main_window.py
tlambert03: still there ...

micromanager_gui/multid_widget.py
```diff
@@ -185,6 +206,8 @@ def set_multi_d_acq_dir(self):
         self.dir_lineEdit.setText(self.save_dir)
         self.parent_path = Path(self.save_dir)
 
+        # self.toggle_run_btn()
```
tlambert03, line 209: remove comments

micromanager_gui/multid_widget.py
```diff
@@ -249,9 +272,25 @@ def _on_run_clicked(self):
...
+            else:
+                experiment = MDASequence(**self._get_state_dict())
+                self._mmc.run_mda(experiment)  # run the MDA experiment asynchronously
+                return
+
+        else:
+            experiment = MDASequence(**self._get_state_dict())
+            self._mmc.run_mda(experiment)  # run the MDA experiment asynchronously
+            return
+
```
tlambert03, line 293: refactor this to remove the code duplication... also, you don't need to use if/else if you're going to raise an error.  so this can just be something like
```python
if self.save_groupBox.isChecked() and some_other_condition_is_not_met:
    raise ValueError ('select a filename and a valid directory.')

experiment = MDASequence(**self._get_state_dict())
self._mmc.run_mda(experiment)  # run the MDA experiment asynchronously
return
```

micromanager_gui/multid_widget.py
```diff
@@ -249,9 +272,25 @@ def _on_run_clicked(self):
...
-        self._mmc.run_mda(experiment)  # run the MDA experiment asynchronously
-        return
+        if self.save_groupBox.isChecked(): 
+
+            if self.fname_lineEdit.text() == '' or \
+                (self.dir_lineEdit.text() == '' or \
+                    not Path.is_dir(Path(self.dir_lineEdit.text()))
+                    ):
+                        raise ValueError ('select a filename and a valid directory.')
+                        print('select a filename and a valid directory.')
```
tlambert03, line 282: no print statements

micromanager_gui/main_window.py
```diff
@@ -152,19 +175,126 @@ def add_frame_explorer(self, name, array):
...
+                    n = n[::-1]
+                    fname = fname.replace(n, '_' + '{0:03}'.format(int(n)))
+                    fname = self.get_filename(fname,list_dir)
+                else:
+                    fname = fname + '_000'
+                    fname = self.get_filename(fname,list_dir)
+
+
+            #TODO: save also metadata
+            self.viewer.layers[str(active_layer)].save(str(save_path / fname))
```
tlambert03, line 272:
```suggestion
            active_layer.save(str(save_path / fname))
```
i think we've seen this before... but want to mention again: there's no need to "re-get" the layer from `viewer.layers` here.  `active_layer` _IS_ the layer... so you're basically just saying "give me the layer from `viewer.layers` that has the same name as `active_layer`" (which will always just _be_ `active_layer`)

micromanager_gui/main_window.py
```diff
@@ -152,19 +175,126 @@ def add_frame_explorer(self, name, array):
...
+
+            """update filename in mda.fname_lineEdit for the next aquisition."""
```
tlambert03, line 274: use `#` type comment

tlambert03: also ... the `list_dir` variable will go out of scope at the end of this function (i.e. it's not `self.list_dir`) ... so you're not actually updating anything for the next acquisition here.  In any case though ... getting a list of files in a directory is a fast operation, so it's not worth maintaining a separate internal list that has the potential to go out of sync with the folder.  just check it every time

micromanager_gui/main_window.py
```diff
@@ -152,19 +175,126 @@ def add_frame_explorer(self, name, array):
...
+            list_dir.append(fname + '.tif')
+            fname = self.get_filename(fname,list_dir)
+            self.mda.fname_lineEdit.setText(fname)
+
+        """reactivate gui when mda finishes."""
```
tlambert03, line 279: `#` comment

micromanager_gui/__main__.py
```diff
@@ -4,5 +4,5 @@
 viewer = napari.Viewer()
 win = MainWindow(viewer)
 viewer.window.add_dock_widget(win, area="right", allowed_areas=["left", "right"])
-viewer.window._qt_window.showFullScreen()
+# viewer.window._qt_window.showFullScreen()
```
tlambert03, line 7: no commented out code... just delete it

micromanager_gui/main_window.py
```diff
@@ -152,19 +175,126 @@ def add_frame_explorer(self, name, array):
         except KeyError:
             self.viewer.add_image(array, name=layer_name)
 
-    # TO DO: add the file name form the save box
+
+    def _on_mda_started(self, sequence: useq.MDASequence):
+        """"create temp folder when mda starts."""
+
+        self.temp_folder = tempfile.TemporaryDirectory(None, str(sequence.uid))
```
tlambert03, line 182: i don't see where this is ever cleaned up with `self.temp_folder.cleanup()`.  It's important that running an experiment has no "side effects" (leaving little things here and there).  So _on_mda_finished needs to call the `cleanup` method described in the [docstring](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory).

micromanager_gui/main_window.py
```diff
@@ -152,19 +175,126 @@ def add_frame_explorer(self, name, array):
...
+
+        """blok gui when mda starts."""
```
tlambert03, line 184: `#` style comment, typo: blok -> block

micromanager_gui/main_window.py
```diff
@@ -152,19 +175,126 @@ def add_frame_explorer(self, name, array):
...
+        self.mda.disable_mda_groupbox()
+        self.disable_gui()
+
+    
+    def get_filename(self, fname, list_dir):
```
tlambert03, line 189: I see now that you've gathered `list_dir` to check for the existence of files...   pathlib already has an [`exists`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.exists) function for this... so no need to hold on to `list_dir` just for that.  Just build the string you would like to check, and then use either `os.path.exists('some_name')` or use `Path('some_name').exists()` ... don't check to see if it is in the list

micromanager_gui/main_window.py
fdrgsp: I read about that but I checked and the temp folder that I create is deleted automatically.

I will add the cleanup() once the mda is finished.

micromanager_gui/multid_widget.py
fdrgsp: Thanks!

micromanager_gui/main_window.py
fdrgsp: I added:

```
if hasattr(self, 'temp_folder'):
     savefile = Path(self.temp_folder.name) / image_name
     tifffile.tifffile.imsave(str(savefile), image)
```

but there is still an error. Looking in the error details:

```
 File "/Users/runner/work/napari-micromanager/napari-micromanager/micromanager_gui/main_window.py", line 459, in change_objective
32
    raise ValueError (
33
ValueError: MAGNIFICATION NOT SET, STORE OBJECTIVES NAME STARTING WITH e.g. 100X or 100x.
```

tlambert03: that's because you changed it to an error [in this PR](https://github.com/tlambert03/napari-micromanager/pull/13/files#diff-649c4a93f80a0ec181a3ef9ae704d234cd8d7745d59dba35220e94166ed3f428R459) :joy:

I think a value error is a bit strict there.  I don't think we want to make any assumptions about the config file, like "you must store objectives with a specific name".  I'd go back to a print statement and put that whole thing on the todo list for another PR

fdrgsp: Yes but since the code is doing what it is supposed to do, shouldn't it be considered as a "non error"?

tlambert03: maybe it's working for your config but not for the demo config provided by micromanager? :)

tlambert03: basically... for some reason or another, it _is_ erroring here.  so the challenge (and the whole purpose/benefit of having tests here ...) is to figure out what assumption or code caused the test to fail, where it was passing before.  Sometimes you'll find that the test itself is in error, other times you'll find that it caught something you hadn't anticipated.
In this case, I suspect that you typically develop with one very specific demo config, and it works for that case... but that's not what is being tested here.

fdrgsp: > maybe it's working for your config but not for the demo config provided by micromanager? :)

Yes it works only with the demo_config I created just because from the names of the objectives, the code is able to extract the magnification and so calculate the pixel size.

tlambert03: right... and I'm saying that's a bit _too_ strict.  In other words, we can't tell someone using this program their config file, which is totally valid and works in micromanger, is no longer valid here... and raise a `ValueError`.  You can certainly _look_ for some sort of convention and print a warning or something if you don't find it (but even there I think we should try to figure out if it can be done without using a special naming convention)... But you can't actually raise an exception and stop everything.

In any case, the point here is that you need to revert line 458 either back to a print statement, or use 
```python
import warnings

warnings.warn('some warning')
```

anytime you use `raise` ... even if it's `raise Warning` ... you're going to completely stop execution of the program.
https://docs.python.org/3/tutorial/errors.html#raising-exceptions

fdrgsp: Thanks!!!

I was reading about that and I changed to "Warning" to see what was happening...I'll read a bit more about that

PR #15, opened 2021-06-01
add option to save xy positions in a separate file

fdrgsp:


fdrgsp:
@tlambert03 ready for review

PR #16, opened 2021-06-02
Add options for individual positions and individual  channels

fdrgsp:
- Add option to save individual positions as separate files;
- Add option to split channels in separate layer during mda;


fdrgsp:
@tlambert03 ready for review

PR #17, opened 2021-06-03
MDA and Sample Explorer tabs - work in progress

fdrgsp:
- Work on MDA
- Add sample explorer functionality using useq (It needs info about the pixel size, so the micromanager "demo_config.cfg" has
   to be loaded).


fdrgsp:
@tlambert03 ready for review

micromanager_gui/main_window.py
```diff
@@ -168,7 +168,7 @@ def get_event(viewer, event):
             if not self._mmc.getPixelSizeUm() > 0:
                 x = None
                 y = None
-                raise ValueError ('PIXEL SIZE NOT SET.')
+                raise Warning ('PIXEL SIZE NOT SET.')
```
tlambert03, line 171: don't `raise Warning`.

If your program _can_ continue, but you'd like the user to know something unexpected happened, then use `import warnings;  warnings.warn("some message")`

If you _cannot_ continue the execution then use `raise`, but try to find the best corresponding [exception type](https://docs.python.org/3/library/exceptions.html#base-classes)

tlambert03:
Hi @fdrgsp, I merged in the master branch that has a bunch of tests, and it exposed one bug in this code here that needs to be fixed before merge.  whenever a zstack is acquired, I get the following stack trace, which indicates a bug in `_save_mda_acq` at ~ main_window.py:329: 


```pytb
  File "/Users/talley/Dropbox (HMS)/Python/forks/napari-micromanager/micromanager_gui/main_window.py", line 283, in _on_mda_finished
    self._save_mda_acq(sequence, meta)
  File "/Users/talley/Dropbox (HMS)/Python/forks/napari-micromanager/micromanager_gui/main_window.py", line 329, in _save_mda_acq
    tifffile.imsave(
  File "/Users/talley/miniconda3/envs/napdev/lib/python3.9/site-packages/tifffile/tifffile.py", line 891, in imwrite
    result = tif.write(data, shape, dtype, **kwargs)
  File "/Users/talley/miniconda3/envs/napdev/lib/python3.9/site-packages/tifffile/tifffile.py", line 1694, in write
    ijshape = imagej_shape(
  File "/Users/talley/miniconda3/envs/napdev/lib/python3.9/site-packages/tifffile/tifffile.py", line 14446, in imagej_shape
    raise ValueError('ImageJ hyperstack is not a grayscale image')
ValueError: ImageJ hyperstack is not a grayscale image
```

 you should pull these changes to your local branch, and then you can run `pytest` to see whether you've fixed them

micromanager_gui/main_window.py
```diff
@@ -139,36 +141,82 @@ def __init__(self, viewer: napari.viewer.Viewer, remote=True):
         self.bit_comboBox.currentIndexChanged.connect(self.bit_changed)
         self.bin_comboBox.currentIndexChanged.connect(self.bin_changed)
 
+        @self.viewer.mouse_drag_callbacks.append
```
tlambert03, line 144: another thing that would be desirable, would be to move as much of the sample-explorer specific logic, into the sample-explorer widget. In the interest of keeping things modular, it's best to limit cross-widget communication to the bare minimum.

It looks to me like this whole mouse callback is really just for the sample explorer, right? If so, it should be in the sample explorer file. If you need to pass the viewer to the sample-explorer constructor, that's fine

micromanager_gui/main_window.py
```diff
@@ -231,13 +475,16 @@ def _refresh_camera_options(self):
...
     def _refresh_objective_options(self):
         if "Objective" in self._mmc.getLoadedDevices():
+            self.objective_comboBox.clear()
             self.objective_comboBox.addItems(self._mmc.getStateLabels("Objective"))
 
     def _refresh_channel_list(self):
         if "Channel" in self._mmc.getAvailableConfigGroups():
+            self.snap_channel_comboBox.clear()
+            self.mda.clear_channel()
+            self.explorer.clear_channel()
```
tlambert03, line 485: same thing here... can the event that should trigger self.explorer to clear it's channels, be connected in that widget, rather than here?

micromanager_gui/main_window.py
```diff
@@ -261,6 +508,8 @@ def bin_changed(self):
     def _on_xy_stage_position_changed(self, name, x, y):
         self.x_lineEdit.setText(f"{x:.1f}")
         self.y_lineEdit.setText(f"{y:.1f}")
+        self.explorer.x_lineEdit.setText(str(x))
+        self.explorer.y_lineEdit.setText(str(y))
```
tlambert03, line 512: ... and here

micromanager_gui/main_window.py
```diff
@@ -516,22 +518,32 @@ def _on_stage_position_changed(self, name, value):
             self.z_lineEdit.setText(f"{value:.1f}")
 
     def stage_x_left(self):
-        self._mmc.setRelPosition(dx=-float(self.xy_step_size_SpinBox.value()))
+        self._mmc.setRelativeXYPosition(-float(self.xy_step_size_SpinBox.value()), 0.0)
```
tlambert03, line 521: thanks!  forgot about that change

fdrgsp:
> Hi @fdrgsp, I merged in the master branch that has a bunch of tests, and it exposed one bug in this code here that needs to be fixed before merge. whenever a zstack is acquired, I get the following stack trace, which indicates a bug in `_save_mda_acq` at ~ main_window.py:329:
> 
> ```
> Traceback (most recent call last):
>   File "/Users/talley/Dropbox (HMS)/Python/forks/napari-micromanager/micromanager_gui/main_window.py", line 283, in _on_mda_finished
>     self._save_mda_acq(sequence, meta)
>   File "/Users/talley/Dropbox (HMS)/Python/forks/napari-micromanager/micromanager_gui/main_window.py", line 329, in _save_mda_acq
>     tifffile.imsave(
>   File "/Users/talley/miniconda3/envs/napdev/lib/python3.9/site-packages/tifffile/tifffile.py", line 891, in imwrite
>     result = tif.write(data, shape, dtype, **kwargs)
>   File "/Users/talley/miniconda3/envs/napdev/lib/python3.9/site-packages/tifffile/tifffile.py", line 1694, in write
>     ijshape = imagej_shape(
>   File "/Users/talley/miniconda3/envs/napdev/lib/python3.9/site-packages/tifffile/tifffile.py", line 14446, in imagej_shape
>     raise ValueError('ImageJ hyperstack is not a grayscale image')
> ValueError: ImageJ hyperstack is not a grayscale image
> ```
> 
> you should pull these changes to your local branch, and then you can run `pytest` to see whether you've fixed them

Hey @tlambert03, I think I found the issue...basically the tifffile kwarg `imagej=True` is supporting stacks up 5 dimensions. So I guess we have 2 options:
- not caring about the imagej compatibility of the stacks saved
- or save each position in separate files all the time to reduce the stack size to max 5 dimensions.

side note:  even when there is only 1 position, the stack we generate is like: `tpzcxy ->(2,1,3,2,512,512)`

What do you suggest?

tlambert03:
hmmm.  well, the ImageJ compatibility is nice, but I don't think we necessarily want to limit ourselves based on it.  Perhaps you could do
```
data = data.squeeze()  # remove singleton dimensions
imsave(..., imagej=data.ndim <=5)
```
?

micromanager_gui/main_window.py
fdrgsp: @tlambert03 I moved everything related to the explorer widget into sample_explorer.py (move all explorer features in explorer widget: https://github.com/tlambert03/napari-micromanager/pull/17/commits/22e3419e34bc2d041069459677054783179bdb28)

tlambert03: looks good thanks.  but something has broken in the meantime: the `test_main_window` test is now failing... any ideas?

fdrgsp: I've been trying to understand why but I'm not sure...it seems that with the current settings in `test_main_window` there won't be any layer in the viewer...
Can it be because `SEQUENCE_META` is not defined in the tests?
```
def _on_mda_frame(self, image: np.ndarray, event: useq.MDAEvent):
        seq = event.sequence
        meta = self.mda.SEQUENCE_META.get(seq, {})

        if meta.get("mode") == "mda":
          ...
```

tlambert03: not sure yet, but I can say that it happened in 22e3419 (697c7e2e8c5440475339e1c4ab5c492aa9a5f168 was fine)

fdrgsp: @tlambert03 Ok I think is fixed. The test was missing the `SEQUENCE_META` info.

tests/test_main_window.py
```diff
@@ -45,6 +45,15 @@ def main_window(qtbot, request):
     win = MainWindow(viewer=viewer, remote=request.param == "remote")
     win._mmc.loadSystemConfiguration("demo")
 
+    win.mda.SEQUENCE_META = {
```
tlambert03, line 48: hmm, this feels like we're just making the test work with the new code. but not necessarily considering whether the test was actually "correct" already.

I need to take a closer look, but my immediate feeling here is that we've taken a method that takes a clean object (a useq MDAevent) and has a predictable response, and turned it into something that will only work correctly if some specific state has already been satisfied on the object (that's what this test "fix" is doing... it's making sure that the test has the state it needs pass). I'd like to take another look at why the mda_frame method now _also_ needs to know something about the SEQUENCE_META object.  will do so soon

fdrgsp: Sure! Maybe instead of:
```
def _on_mda_frame(self, image: np.ndarray, event: useq.MDAEvent):
        seq = event.sequence
        meta = self.mda.SEQUENCE_META.get(seq, {})

        if meta.get("mode") == "mda":
            ...
```
we can change it with:

           `if not "mode" in meta`

or something similar...not sure...

micromanager_gui/multid_widget.py
```diff
@@ -15,6 +17,16 @@
 ICONS = Path(__file__).parent / "icons"
 
 
+@dataclass
+class SequenceMeta:
+    mode: Literal["mda"] | Literal["explorer"] = ""
```
tlambert03, line 22: hey @fdrgsp, can you comment on why you wanted the default to be the empty string here?
This type annotation is no longer accurate (and `test_main_window` is now failing as a result)...  The type `Literal["mda"] | Literal["explorer"]` means "this value should either be the string `"mda"` or the string `"explorer"` ... but then we set it to neither of those.  thoughts?

fdrgsp: Hey @tlambert03, I've seen the tests failed and I was working on it...but basically if `"mode"` is set to `"mda"` as default, when using the explorer both `_on_explorer_frame ` and `_on_mda_frame` are executed.

tlambert03: ah, k!

micromanager_gui/prop_browser.py
```diff
@@ -226,9 +226,9 @@ def _on_le_change(self, e):
 
 
 if __name__ == "__main__":
-    from micromanager_gui._core._mmcore_plus import MMCorePlus
+    from pymmcore_plus import CMMCorePlus
```
ianhi, line 230: Maybe should be `RemoteMMCore` ?

tlambert03: perhaps, but lets actually leave this out of this PR altogether.  @fdrgsp  can you revert any changes made to the prop_browser?  let's deal with those in a follow up PR

tlambert03:
thanks for all the work @fdrgsp!  I'm going to merge this, we can work on test coverage next

PR #20, opened 2021-07-06
Allow exposure to be a float + connect to exposure changed signal

ianhi:
Previously if you also had a separate client to `pymmcore-remote` that changed the exposure then the gui would not update and then would overwrite the  value.

There is a similar issue with `channels` (perhaps config groups generally) but I'm struggling with figuring out the `gets` and sets`. I would open an issue about this by it seems that issues are disabled on this repo currently.

### DIFF
--- a/micromanager_gui/main_window.py
+++ b/micromanager_gui/main_window.py
@@ -57,7 +57,7 @@ class _MainUI:
     multid_tab: QtW.QWidget
     snap_channel_groupBox: QtW.QGroupBox
     snap_channel_comboBox: QtW.QComboBox
-    exp_spinBox: QtW.QSpinBox
+    exp_spinBox: QtW.QDoubleSpinBox
     snap_Button: QtW.QPushButton
     live_Button: QtW.QPushButton
     max_val_lineEdit: QtW.QLineEdit
--- a/micromanager_gui/main_window.py
+++ b/micromanager_gui/main_window.py
@@ -343,7 +343,7 @@ class MainWindow(QtW.QWidget, _MainUI):
 
     def snap(self):
         self.stop_live()
-        self._mmc.setExposure(int(self.exp_spinBox.value()))
+        self._mmc.setExposure(self.exp_spinBox.value())
         self._mmc.snapImage()
         self.update_viewer(self._mmc.getImage())
 
--- a/micromanager_gui/main_window.py
+++ b/micromanager_gui/main_window.py
@@ -348,7 +348,7 @@ class MainWindow(QtW.QWidget, _MainUI):
         self.update_viewer(self._mmc.getImage())
 
     def start_live(self):
-        self._mmc.startContinuousSequenceAcquisition(int(self.exp_spinBox.value()))
+        self._mmc.startContinuousSequenceAcquisition(self.exp_spinBox.value())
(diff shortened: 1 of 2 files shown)


tlambert03 (approved):
lovely thanks.

as you've probably noticed, there's a lot of gui-state independence from the mmcore state in the `main_window` (i.e. it loads when we load the config, but isn't connected well to the mmcore signals).  That's not by design, so thanks for getting us going in the right direction here!

micromanager_gui/main_window.py
```diff
@@ -342,15 +344,15 @@ def update_viewer(self, data=None):
...
         self._mmc.snapImage()
         self.update_viewer(self._mmc.getImage())
 
     def start_live(self):
-        self._mmc.startContinuousSequenceAcquisition(int(self.exp_spinBox.value()))
+        self._mmc.startContinuousSequenceAcquisition(self.exp_spinBox.value())
         self.streaming_timer = QTimer()
         self.streaming_timer.timeout.connect(self.update_viewer)
-        self.streaming_timer.start(int(self.exp_spinBox.value()))
+        self.streaming_timer.start(ceil(self.exp_spinBox.value()))
```
ianhi, line 355: I'm not sure that this is the best thing here. Maybe `int` is better? Or we could also try `floor`

tlambert03: `int` and `floor` should be identical for positive numbers right? I don't have a strong feeling here.  exposure is in milliseconds right?

ianhi: Yes identical indeed. Yeah, exposure is in milliseconds. So the question is: is it worse to potentially let the buffer of images grow steadily (admittedly very slowly). Or to sometimes skip a frame when the viewer gets ahead of the acquisition stream. 

I suppose `round` avoids both worst cases but also will change behavior based on exposure time. 

--- 
I'm also not sure that this matters terribly much as the difference between these will never be more than a millisecond. So the potential worst cases are exposure times of `1.01` ms and `1.99`

tlambert03: > So the question is: is it worse to potentially let the buffer of images grow steadily (admittedly very slowly). Or to sometimes skip a frame when the viewer gets ahead of the acquisition stream.

the circular buffer is on the micromanager side, so I feel like it's only the second issue that's at play here (and it applies whether the viewer gets ahead or behind of the acquisition stream).  And I agree that the worst case scenario is `1.01` and `1.99` ... _however_: at that point, I don't think the human visual system is capable of perceiving the periodic asynchrony :)

so, either one is fine with me.  are you otherwise done on this PR?

ianhi: I went back to `int`. After that I think I'm done here. I have some follow ups with the remaining signals from pymmcore-plus that are still missing (in particular the channels) but those may take a bit more work

ianhi: Also @tlambert03 bump of the idea of enabling issues? There are a few other follow things I noticed that it would be good to not lose track of.

tlambert03: > Also @tlambert03 bump of the idea of enabling issues?

sorry!  I totally missed that.  enabled

Issue #26, opened 2021-07-27
Use `channelGroup` to determine the config group used for the channel


ianhi:
There is `core.getChannelGroup` which should be used instead of hardcoding `"Channel"` https://github.com/tlambert03/napari-micromanager/blob/57ee6672aad52e171058376f2a115305f2dabc84/micromanager_gui/main_window.py#L237-L239

However, we should keep the hardcoded `Channel` as a fallback because it's not clear if it's even possible to set this via a config file (and micromanager doesn't seem to save it into the config) see https://forum.image.sc/t/how-to-set-channelgroup-in-config-file/55746

tlambert03:
agreed!

Issue #27, opened 2021-07-27
Development tips?


ianhi:
Any tips+tricks for developing a napari plugin? In particular is there an easy way to start up napari and immediately load the plugin? Every time I want to test something I am opening napari from the terminal, loading the plugin and then loading the demo config.

tlambert03:
> In particular is there an easy way to start up napari and immediately load the plugin?

tlambert03:
and here's a script for adding a plugin dock widget to an existing viewer via a script:

```py
import napari


v = napari.Viewer()
dw, main_window = v.window.add_plugin_dock_widget('micromanager')

core = main_window._mmc
core.loadSystemConfiguration('demo')

napari.run()
```

Issue #21, opened 2021-07-07
Properly connect gui state and core state


tlambert03:
as observed by @ianhi  in #20, the gui was written before there was a good way to receive callbacks from the core object.  We need to go back and use event-based gui-state updates instead of on-demand updates.

ianhi:
> to go back and use event-based gui-state updates instead of on-demand updates.

Is there a canonical place where state should live in a qt application? Or resources to read up on the reasonable flow of data

So when you say "instead of on-demand updates" does that mean that things like this shold be written differently?

https://github.com/tlambert03/napari-micromanager/blob/96a8a0cafeaac840fdf6a096b2b9cf4006554569/micromanager_gui/main_window.py#L344-L346

tlambert03:
> Is there a canonical place where state should live in a qt application? Or resources to read up on the reasonable flow of data

In a basic QWidget-based setup (as opposed to a more abstract [model/view](https://doc.qt.io/qt-5/model-view-programming.html) setup) , the widget itself holds state.  For instance (as you know) a QComboBox has a [`currentText`](https://doc.qt.io/qt-5/qcombobox.html#currentText-prop) property that you can get/set, but it's not intrinsically linked elsewhere.

Looking at the code again now, I see a lot fewer "on-demand" updates than I remember, but what I mostly meant were methods that we manually call to _query_ the core and some gui state, rather than gui state that is directly listening to a signal emitted by the core object.   For instance, all the `_refresh_x` methods, that get called by `_on_system_configuration_loaded`... I _think_ (at least, it's worth exploring) that we should be connecting directly to the core [`onPropertyChanged`](https://valelab4.ucsf.edu/~MM/doc/MMCore/html/class_core_callback.html#a83d0973bfbd2a237ede596333b7edea7) event.    So, after https://github.com/tlambert03/pymmcore-plus/pull/14 is merged:

```python
In [1]: from pymmcore_plus import CMMCorePlus

In [2]: mmc = CMMCorePlus()

# just print anytime a property changes
In [3]: mmc.propertyChanged.connect(lambda *props: print(props))

In [4]: mmc.loadSystemConfiguration()
('Core', 'Initialize', '0')
('Core', 'Initialize', '1')
('Core', 'Camera', 'Camera')
('Core', 'Shutter', 'Shutter')
('Core', 'Focus', 'Z')
('Core', 'AutoShutter', '1')
('Objective', 'State', '1')
[...]

PR #28, opened 2021-07-27
Add callbacks to deal with setting channel

ianhi:
This PR adds two things
1. The development helper script from #27 and added a brief description to the readme.
2. bi-directional callbacks for when the channel is changed (also use the channelGroup when possible)

Closes: https://github.com/tlambert03/napari-micromanager/issues/26

Gets at part of #21, I didn't use the declarative mapping here because I think that this is the only config group that currently has an effect on the GUI. However, your proposed mapping will be helpful for the device properties.

### DIFF
--- a/micromanager_gui/main_window.py
+++ b/micromanager_gui/main_window.py
@@ -116,6 +116,7 @@ class MainWindow(QtW.QWidget, _MainUI):
         sig.stagePositionChanged.connect(self._on_stage_position_changed)
         sig.exposureChanged.connect(self._on_exp_change)
         sig.frameReady.connect(self._on_mda_frame)
+        sig.channelGroupChanged.connect(self._refresh_channel_list)
 
         # connect explorer
         self.explorer.new_frame.connect(self.add_frame_explorer)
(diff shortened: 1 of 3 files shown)


ianhi:
Add a pre-commit.ci workflow here as well?

micromanager_gui/main_window.py
```diff
@@ -258,6 +271,21 @@ def bin_changed(self):
             cd = self._mmc.getCameraDevice()
             self._mmc.setProperty(cd, "Binning", bins)
 
+    def _get_channel_group(self) -> Union[str, None]:
+        """
+        Get channelGroup falling back to Channel if not set, also
+        check that this is an availableConfigGroup.
+        """
+        chan_group = self._mmc.getChannelGroup()
```
ianhi, line 279: This is a low cost call - probably the same as accessing an attribute on self https://github.com/micro-manager/mmCoreAndDevices/blob/97f82b9ef02788d8b6e9aa70283eb10c7c645a6e/MMCore/MMCore.cpp#L3470-L3474

ianhi:
I ended up moving towards a strategy of constantly re-checking the channelgroup from the core instead of grabbing it early and holding it as state on the main-window widget. When I tried the latter I ran into some trickiness with timing of callbacks and the definition of the variable, as well I think it's probably better to regard the core as the fundamental source of truth.

tlambert03:
added

tlambert03 (approved):
thanks @ianhi, this looks good.  anything else?

ianhi:
Nope I think that's all for this one. More to come in the future

Issue #33, opened 2021-08-13
`_on_system_configuration_loaded` should not modify system state


ianhi:
Noticed today that when I opened napari-micromanager (manually called `_on_system_configuration_loaded` because it was already loaded remotely) that the objective position changed from where it already was. This seems to be because the objective_combobox was populated with a list default order but didn't check the current state of the objectives so it ended up calling `change_objective`.

I have two proposed improvements:

1. [ ] At the end of `init` the system state should be queried  to account for the config already being loaded
2. [ ] `refresh_objective` and friends should take care to not accidentally change the current state via the callbacks on their widgets.

tlambert03:
definitely anything you want to submit that protects potential damage is very welcome and high priority

PR #35, opened 2021-08-20
Add safety checks to all `_refresh` methods

ianhi:
This should prevent accidental changing hardware state when updating the available options. I also added refreshing the options at the end of init in case another remote has already loaded a config.

Finally in e6195d3 I removed some hardcoded setting of bin depth which didn't seem approriate

Should fix: #33


micromanager_gui/main_window.py
```diff
@@ -112,9 +112,7 @@ def __init__(self, viewer: napari.viewer.Viewer, remote=True):
         # to core may outlive the lifetime of this particular widget.
         sig.sequenceStarted.connect(self._on_mda_started)
         sig.sequenceFinished.connect(self._on_mda_finished)
-        sig.sequenceFinished.connect(
-            self._refresh_options
-        )  # why when acq is finished?
+        sig.sequenceFinished.connect(self._refresh_options)  # why when acq is finished?
```
ianhi, line 115: following up on this comment - why is this happening? I think if the signals are set up correctly then the gui should stay up to date through the mda

ianhi:
working on adding a test

micromanager_gui/main_window.py
```diff
@@ -245,46 +245,58 @@ def load_cfg(self):
...
-    def _on_system_configuration_loaded(self):
-        self._refresh_camera_options()
-        self._refresh_objective_options()
-        self._refresh_channel_list()
-        self._refresh_positions()
+            with blockSignals(self.snap_channel_comboBox):
+                self.snap_channel_comboBox.addItems(channel_list)
+                self.snap_channel_comboBox.setCurrentText(
+                    self._mmc.getCurrentConfig("Channel")
+                )
```
ianhi, line 287: Could also use `getCurrentConfigFromCache` for faster method, but this should be closer to the ground truth and since refresh should only be called rarely I think it's best to do this.

ianhi:
Test is failing because the demo config in this repo (which I based the test on) is different than the one distributed by micromanager: https://github.com/micro-manager/micro-manager/blob/master/bindist/any-platform/MMConfig_demo.cfg

@tlambert03 what do you think is best w.r.t which demo file to use. My inclination is to have a demo file in this repo used for the tests so that upstream  changes to the demo don't mess with the tests.

ianhi:
@tlambert03 I think this is done now

ianhi:
Bump of this - the `blockSignals` stuff has implications for things like #43, also I have some time to use this on a scope this week and will otherwise need to be basing any improvements off this branch because I don't want to use on a real scope without these safety checks.

tlambert03 (approved):
sorry for the delay @ianhi ... busy week.  I'll just merge this one optimistically and we can iterate as needed.  Thanks for working on this!

ianhi:
Thanks talley!

> sorry for the delay @ianhi ... busy week. 

no worries at all! I figured as much didn't mean to bug

PR #39, opened 2021-09-01
move the xystage back to the starting position when the explorer scan is finished

fdrgsp:


micromanager_gui/explore_sample.py
```diff
@@ -130,6 +133,8 @@ def _on_explorer_frame(self, image: np.ndarray, event: useq.MDAEvent):
         self.viewer.reset_view()
 
     def _on_mda_finished(self, sequence: useq.MDASequence):
+        if self.return_position_x is not None and self.return_position_y is not None:
+            self._mmc.setXYPosition(self.return_position_x, self.return_position_y)
```
ianhi, line 137: I'm worried about this being in `_on_mda_finished` because this function can be triggered from multiple places not after a grid exploration. For example you could do the grid exploration, then set up an MDA independently of that in a totally different position, run it, and then the stage would return to the start position of the initial grid exploration.

One way around this could maybe be to have the concept of `run_once` callbacks in `pymmcore_plus` so that `set_grid` or `start_scan` could register a callback that would run only this one time and not on subsequent MDAs initiated from another process or by the MDA panel.

ianhi (changes requested):
This is a nice feature for grid explore but I think that the current implementation will have a negative side effects on other MDAs.

micromanager_gui/explore_sample.py
fdrgsp: sure, you're right. for now I've added a variable reset to None in `_on_mda_finished` but it might be ok to have in in `pymmcore_plus`.

PR #40, opened 2021-09-01
Use `getLastImage` for live mode

ianhi:
This makes the display significantly less laggy when moving the stage around. Inspired by MMs implementation.

Inspired by https://github.com/micro-manager/micro-manager/blob/4a5d51ea76f89eaa6e74d0d772822701661aa76a/mmstudio/src/main/java/org/micromanager/internal/SnapLiveManager.java#L406

### DIFF
--- a/micromanager_gui/main_window.py
+++ b/micromanager_gui/main_window.py
@@ -385,7 +385,7 @@ class MainWindow(QtW.QWidget, _MainUI):
         #       - are max and min_val_lineEdit updating in live mode?
         if data is None:
             try:
-                data = self._mmc.popNextImage()
+                data = self._mmc.getLastImage()
             except (RuntimeError, IndexError):
                 # circular buffer empty
                 return


tlambert03:
wow!  that's it huh?  amazing.
(I have a lot to learn about micromanager :joy:)

tlambert03 (approved):

ianhi:
It's wild how much complexity they've managed to hide from me as a user. For example they deal with multi camera setups:

>          // We scan over 2*numCameraChannels here because, in multi-camera
         // setups, one camera could be generating images faster than the
         // other(s). Of course, 2x isn't guaranteed to be enough here, either,
         // but it's what we've historically used.

there's tons of stuff like that where you need to really cover everyone, very impressive - also makes me wonder what the appropriate scope for this library is.

tlambert03:
> very impressive - also makes me wonder what the appropriate scope for this library is.

yep, I'm not surprised at all to hear it.  it's a very mature program written by smart people :)

i don't have an answer for the scope here... I'd say it's an experiment that has no particular stopping criterion.  I still dream of some sort of python/C implementation that doesn't require Java, and has access to the rich library of device adapters.  So until that happens somewhere I'm willing to keep chugging along :)

PR #43, opened 2021-09-02
get objective device and check if  there is a configuration group

fdrgsp:
Check if there is an objectives device (e.g. TINosePiece for Nikon microscopes).
- If true, check if there is a group configuration for the objectives. 
   - If there is one, the presets are added to the objective combobox in the gui. 
   -  If there is not a group configuration, the objectives "labels" are used and added to the objective combobox in the gui.
- If false, the objective combobox in the gui will be empty.

https://user-images.githubusercontent.com/70725613/144635258-df2ece66-8d78-4f6e-814a-8fc2770c3055.mov


ianhi:
This will also require modifying tthe code in `change_objective` because that currently searches for the `Objective` device:

https://github.com/tlambert03/napari-micromanager/blob/64cf5c65dd934e76718b8d9d024e5aade2d09e5e/micromanager_gui/main_window.py#L359-L362

It's nice that this adds the safety of lowering the objective which I think would be hard to support when switching to configgroups :(. Is it common for the device drivers to handle that or does it really need to be here?

fdrgsp:
still working on it...

fdrgsp:
I think this should work:

1. if there is not an objective config group but the objective device is included in `OBJ_PTRN`.
2. if there is an objective config group and:
       - the objective config group name is included in `OBJ_PTRN`.
       - the objective config group is defined in micromanager using either the device status (like in the `test_config.cfg` file in the test folder) or the device labels.

ianhi:
@fdrgsp sorry for the mess of the conflict. I think you got there with commits but in general an easier strategy is to use git's rebase functionality (https://git-scm.com/docs/git-rebase). In this case once #35 was merged you would have done:

```bash
git fetch upstream
git rebase upstream/main
# ... deal with the conflicts
git add <conflicting files>
# git rebase --continue 
git push -f
```

micromanager_gui/main_window.py
```diff
@@ -267,13 +272,37 @@ def _refresh_camera_options(self):
...
+                    self.objective_comboBox.clear()
+                    self.objective_comboBox.addItems(
+                        self._mmc.getAvailableConfigs(self.objectives_device)
+                    )
+                    self.set_pixel_size()
+                    return
+
+        for device in self._mmc.getLoadedDevicesOfType(DeviceType.StateDevice):
+            if OBJ_PTRN.match(device):
+                self.objectives_device = device
```
ianhi, line 301: Maybe this could actually live in `pymmocre-plus`? Seems generically useful and also thus far pymmcoreplus has been where the complexity and idiosyncracies of pymmcore has been smoothed over.

ianhi: somehting like this on `CMMCorePlus`

```python
import re
from typing import List, Pattern
from pymmcore_plus import DeviceType
def getObjectiveDevices(self, device_regex: Pattern=None)-> List[str]:
    
    devices = []
    if device_regex is None:
        device_regex = re.compile("(.+)(Nosepiece|Objective|obj)s?", re.IGNORECASE)
    
    for device in self.getLoadedDevicesOfType(DeviceType.StateDevice):
        if device_regex.match(device):
            devices.append(device)
    return devices
```

ianhi: See https://github.com/tlambert03/pymmcore-plus/pull/53 for a proposal

fdrgsp:
I added/changed 3 main methods here:

- `_refresh_objective_options()` -> it uses the new `guessObjectiveDevices()` method in pymmcore-plus to get a list of possible` objective devices`. If there is only one element in the list (only 1 possible objective device), then the objective device is set through the new `_set_objective_device`() method (see below). If there are more elements in the list, it lets you chose the correct objective device through a `magicgui` combobox (`SelectDeviceFromCombobox()` class in `_util.py`)

- `_set_objective_device`() -> checks if there is a configuration group for the objectives. if there is one, it sets both `objectives_device` and objective configuration (`objectives_cfg`) and populate the `objective_comboBox` accordingly. If there is no configuration, it sets the  `objectives_device` and uses the labels to populate the objective combobox.

- set_pixel_size() -> checks if the `pixel size` for the selected objective is already set. If not, it creates and store a new pixel size config.

ianhi:
@fdrgsp can you add some screenshots or a GIF of the usage to this PR's description?

fdrgsp:
@tlambert03 I think you can have a look. I added your suggestions (as described below) as well as a test for the objective device.

- `_refresh_objective_options()` -> it uses the new `guessObjectiveDevices()` method in pymmcore-plus to get a list of possible` objective devices`. 
 
    - If there is only one element in the list (only 1 possible objective device), then the objective device is set through the new `_set_objectives() `method (see below). 
    - If there are more elements in the list, it lets you chose the correct objective device through a `magicgui` combobox (`SelectDeviceFromCombobox()` class in `_util.py`)

- `_set_objectives()  `-> checks through the `_get_objective_device()` method if there is a configuration group for the objectives.
    - if there is one, it sets both `objectives_device` and objective configuration (`objectives_cfg`) and populate the `objective_comboBox` accordingly (through the `_add_objective_to_gui()` method).
    - If there is no configuration, it sets the  `objectives_device` and uses the labels to populate the objective combobox (through the `_add_objective_to_gui()` method).

- `_update_pixel_size()` -> checks if the `pixel size` for the selected objective is already set. If not, it creates and store a new pixel size config.

fdrgsp:
@ianhi Here is an example. It shows that the `objective combobox` is populated with the `objective group configuration presets` (e.g. 10X, 20x, ...) if there is one. Otherwise it will use the `objective device labels`.

https://user-images.githubusercontent.com/70725613/144632275-5bad5685-c22b-4fea-90e4-7e87dec2e752.mov

PR #46, opened 2021-09-03
Small z-position fixes

ianhi:
1. Autofill the z_line-edit when config loaded - otherwise it only updated once you moved the z stage
2. Don't snap when clicking the z up/down buttons (This feels like the correct behavior to me, otherwise you end up stopping live mode if you are already in it.)

### DIFF
--- a/micromanager_gui/main_window.py
+++ b/micromanager_gui/main_window.py
@@ -130,8 +130,6 @@ class MainWindow(QtW.QWidget, _MainUI):
         self.y_down_Button.clicked.connect(self.stage_y_down)
         self.up_Button.clicked.connect(self.stage_z_up)
         self.down_Button.clicked.connect(self.stage_z_down)
-        self.up_Button.clicked.connect(self.snap)
-        self.down_Button.clicked.connect(self.snap)
 
         self.snap_Button.clicked.connect(self.snap)
         self.live_Button.clicked.connect(self.toggle_live)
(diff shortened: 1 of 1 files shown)


micromanager_gui/main_window.py
```diff
@@ -130,8 +130,6 @@ def __init__(self, viewer: napari.viewer.Viewer, remote=True):
         self.y_down_Button.clicked.connect(self.stage_y_down)
         self.up_Button.clicked.connect(self.stage_z_up)
         self.down_Button.clicked.connect(self.stage_z_down)
-        self.up_Button.clicked.connect(self.snap)
```
tlambert03, line 133: we don't have settings yet... but this particular thing here is probably personal preference.  We have a (deltavision) microscope here that has this behavior by default... everyone loves it because it encourages you _not_ to use live mode as you find the proper focal plane (decreases bleaching etc...).  I think @fdrgsp added this to emulate that scope.

It's probably unexpected in many cases though, so I understand wanting to remove it.  would be nice to configure it eventually.

tlambert03 (approved):

micromanager_gui/main_window.py
fdrgsp: Maybe we can add a checkbox to activate this functionality as an option.

ianhi: > we don't have settings yet... but this particular thing here is probably personal preference. 

@tlambert03 is there a generic way for a napari plugin to reuse the settings machinery of napari? Based on my reading of the docs there doesn't seem to be and we'd have to create our own settings panel from scratch? Having settings is definitely starting to seem like an important thing.



> Maybe we can add a checkbox to activate this functionality as an option #50 .

That makes sense to me in this case, but my worry would be that long term there may end up being too many checkboxes and cluttering the UI. Though I don't actually have a clear vision of that, more a general worry.

tlambert03: > @tlambert03 is there a generic way for a napari plugin to reuse the settings machinery of napari? 

not at the moment, but there will be after the new manifest stuff is in place: (https://github.com/napari/napari/issues/2975 and https://github.com/napari/napari/issues/3115)

ianhi:
rebased to pick up test fixes

PR #47, opened 2021-09-03
simplify snap

ianhi:
Now that the callbacks on the GUI are all hooked up snap no longer needs to do expensive things like setting the config group. Also rather than stopping live mode to snap we should disable the ability to snap when live mode is on.

### DIFF
--- a/micromanager_gui/main_window.py
+++ b/micromanager_gui/main_window.py
@@ -416,12 +416,6 @@ class MainWindow(QtW.QWidget, _MainUI):
 
     def snap(self):
         self.stop_live()
-
-        self._mmc.setExposure(self.exp_spinBox.value())
-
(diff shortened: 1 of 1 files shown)


tlambert03:
> Also rather than stopping live mode to snap we should disable the ability to snap when live mode is on.

I don't know about this one.  both elements and micromanager do this (stop live mode if you hit snap)... and I've grown to rely on it and love it.  actually I thought micromanager did it too, no?  Is there a reason you'd rather have it inactive rather than stop live?

ianhi:
Oh interesting. My proposed change is the micromanager behavior: 
image

but I don't really mind either way.

ianhi:
Snap stopping live mode is pretty nice, so i reverted that part.

PR #48, opened 2021-09-03
update tests for newest pymmcore-plus

ianhi:
Accounting for changes to the remote server management in https://github.com/tlambert03/pymmcore-plus/pull/47


tlambert03 (approved):
thanks!

PR #49, opened 2021-09-03
Fully connect exposure signals + less eager exposure modification

ianhi:


ianhi:
@tlambert03 I forgot to mention this as a pretty easy one that also is pretty important. This PR adds three things:
1. The GUI will update if the exposure is changed by another remote (important)
2. Updating the exposure while live mode is on has an effect, otherwise the exposure for live mode stays the same (nice to have)
3. As described in https://github.com/tlambert03/napari-micromanager/pull/49/commits/6a1e2e83f7f7a21b72263da2e15a867d6a16749a the current exposure widget is overeager in changing the exposure which can be annoying when modifying the exposure during live mode. (nice to have/important)

tlambert03 (approved):
🚀

PR #54, opened 2021-09-13
One source of channelgroup + more robust searching

ianhi:
As proposed in #53 Use a single object shared across components for being flexible about names the user may have given to something. Also expand the fallbacks from channelgroup to other common names.

This would supersede #53

----
Maybe this beongs in pymmcore-plus instead? but it's not 100%  to me that it would.


ianhi:
> can you help me understand how this relates to tlambert03/pymmcore-plus#53

While the pymmcore-plus PR deals with objectives and this PR with channelgroups they are both getting at the same idea of having `get___` methods that are robust to user naming. But the big open question if these robust getter methods should live in napari-micromanger or in pymmcore-plus. There are benefits to both ways but we should probably make a choice and then put all the robust getting that we need there. 

This would not close #43, but if we implement this here instead of pymmcore-plus then I think we should merge this, and then rebase #43 to use this robust getter so that we have a clean centralized way of doing the fallbacks. 

I think I lean very slightly towards putting this kind of stuff into pymm+ because I could imagine someone writing their own scripts that are meant to be fairly general and being annoyed that these utilities only exist in the napari-plugin. But the downside there is that it seems like these shouldn't necessarily be just mixed in with the standard pymmcore methods and maybe should get their own `mmc.robust.getChannelGroup` namespace or something like that. 

tldr the important choice is: does this go here, or in pymmcore-plus (I'm 55% for pymm+)

fdrgsp:
I think it can be good to have this in pmmc+. As soon as it is ready and merged, I can add more "get__" methods to this, for example to get the objectives (#43) , maybe the illumination devices (#56) , etc...

PR #56, opened 2021-09-13
add button to create a widget to control illumination devices

fdrgsp:
This might be a starting point to get the illumination devices and create a widget (with magicgui) to control them.

It uses the regex LIGHT_LIST = re.compile("(Intensity|Power)s?", re.IGNORECASE) (similar to #43)

LIGHT_LIST should be expanded.

Still a work in progress.

### DIFF
--- a/micromanager_gui/_illumination.py
+++ b/micromanager_gui/_illumination.py
@@ -1,0 +1,72 @@
+import re
+import sys
+
+from magicgui import magicgui
+from magicgui.widgets import Container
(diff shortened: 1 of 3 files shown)


ianhi:
> It uses the regex LIGHT_LIST = re.compile("(Intensity|Power)s?", re.IGNORECASE) (similar to #43)

love this. Do you have thoughts on something like https://github.com/tlambert03/napari-micromanager/pull/54 which could then also have a `findLightSourceProperties` method?

PR #73, opened 2021-11-04
Prevent from manual editing the cells from the property browser table

fdrgsp:
- The "property" column in the Property Browser table contains the `device_name` and the `device_property` that can be changed using the widgets in the "value" column. The "property" column is currently editable (you can change/delete the content of the cell), I think it can be good to prevent that from happening (`self.native.setEditTriggers(QtWidgets.QTableWidget.NoEditTriggers)`.

- set the `min_width` of the Property Browser table

- hide the `verticalHeader()` of the Property Browser table (row number)


tlambert03:
@fdrgsp ... _please, please_ provide a brief sentence or two explaining your rationale whenever you open a PR, even if it feels self-explanatory.  As is, I'm left to wonder "is this obvious? are these read-only anyway? ... or did he maybe come across a case where editing one that was theoretically editable didn't work when he tried?"

fdrgsp:
sorry @tlambert03, description added.

tlambert03 (approved):

PR #76, opened 2021-11-24
Use new `getChannelGroup getter from pymmc+

ianhi:
Follow up to https://github.com/tlambert03/napari-micromanager/pull/54

Also:

- Expanded the functionality of `blockSignals` to take a list of widgets.
- Be more careful when loading a config that we don't accidentally set properties when `clear`ing the spinboxes


tlambert03 (approved):
LGTM

PR #79, opened 2021-11-30
Update exposure widget when refreshing camera

ianhi:
Otherwise the displayed value may not match up with the true exposure


tlambert03 (approved):
Thanks!

ianhi:
while looking at this I noticed that we aren't doing the kind thing of remembering what the user set the exposure to on a per channel basis. This is a behavior I think MM gets really right and would like to have the ability to replicate here. 

So i added a new `ExposureCache` object that the main window can use to remember what the user last set the exposure to for any given channel.

In the case where the user set the exposure in the config group (something they shouldn't do) it will not override with the most recent value. I could see this going either way but I think that sticking to config value is the safest option.

**Before:**
Only one exposure value - applied to all channels
exp-no-cache


**After:**
exp-cached

ianhi:
Unfortunately I'm really struggling to get the tests to pass. If I execute the test manually via the UI it all behaves as expected, but I can't seem to get it to pass on `[remote]` only on `[local]` so I suspect something is wonky with signal timing, but I can't pin it down.

ianhi:
Final thought: This may be another case of something that should live in pymmcore-plus but again never fully sure and it was easier to test out here first.

tlambert03:
> while looking at this I noticed that we aren't doing the kind thing of remembering what the user set the exposure to on a per channel basis. This is a behavior I think MM gets really right and would like to have the ability to replicate here.

yep agreed, good feature. Did you happen to check how MM implements this?

ianhi:
Start here:

https://github.com/micro-manager/micro-manager/blob/225a28c69a0a04305f85ca580b5722a2d0c6f0de/mmstudio/src/main/java/org/micromanager/internal/dialogs/AcqControlDlg.java#L1020-L1038

Uses these get/set channel exposure times
https://github.com/micro-manager/micro-manager/blob/225a28c69a0a04305f85ca580b5722a2d0c6f0de/mmstudio/src/main/java/org/micromanager/internal/DefaultApplication.java#L134-L166

which do caching via these functions: https://github.com/micro-manager/micro-manager/blob/225a28c69a0a04305f85ca580b5722a2d0c6f0de/mmstudio/src/main/java/org/micromanager/internal/DefaultApplication.java#L161-L172

I think the one difference in behavior is that if you have a channel setting that includes exposure as one of the properties and you:

1. Go to a channel (DAPI)
2. Manually set the exposure to a value different than the config group
3. switch channels
4. back to DAPI

Then the exposure will be what you last set it to, and the channel config drop down menu will go blank the microscope state no longer exactly matches the config group. Whereas here the config group setting will be restored. I can see arguments for both ways here and I think that this is something that should ultimately be togggleable via a settings file.

tests/test_main_window.py
```diff
@@ -27,6 +27,44 @@
...
+def test_exposure_changing(qtbot: "QtBot", main_window: MainWindow):
+    wait = 500
+    mmc = main_window._mmc
+    main_window.snap_channel_comboBox.setCurrentText("DAPI")
+    qtbot.wait_signal(main_window.snap_channel_comboBox.currentTextChanged)
+    qtbot.wait_signals([mmc.events.configGroupChanged, mmc.events.exposureChanged])
+
+    # Wait signals got this working on local but not remote?
+    # although it behaves fine as a user.
+    qtbot.wait(wait)
```
ianhi, line 39: any thoughts on how to make this work on remote? works from a user perspective but somehow not here?

ianhi: i think i fixed it. I had to use a dummy context manager on local in order to avoid this error:

```
xceptions caught in Qt event loop:
________________________________________________________________________________
  File "/home/ian/mambaforge/envs/pymm/lib/python3.9/site-packages/pytestqt/wait_signal.py", line 58, in _quit_loop_by_timeout
    self._cleanup()
  File "/home/ian/mambaforge/envs/pymm/lib/python3.9/site-packages/pytestqt/wait_signal.py", line 572, in _cleanup
    super()._cleanup()
  File "/home/ian/mambaforge/envs/pymm/lib/python3.9/site-packages/pytestqt/wait_signal.py", line 64, in _cleanup
    self._timeout_message = self._get_timeout_error_message()
  File "/home/ian/mambaforge/envs/pymm/lib/python3.9/site-packages/pytestqt/wait_signal.py", line 344, in _get_timeout_error_message
    error_message = self._get_expected_and_actual_signals_message()
  File "/home/ian/mambaforge/envs/pymm/lib/python3.9/site-packages/pytestqt/wait_signal.py", line 519, in _get_expected_and_actual_signals_message
    self._get_signal_string_representation_for_index(missing_signal_index)
  File "/home/ian/mambaforge/envs/pymm/lib/python3.9/site-packages/pytestqt/wait_signal.py", line 555, in _get_signal_string_representation_for_index
    signal_str_repr = self._signal_names[signal]
KeyError: <SignalInstance 'exposureChanged' on <pymmcore_plus.core._signals._CMMCoreSignaler object at 0x7fc9a9d3f9d0>>
________________________________________________________________________________
-------------------------------------------------------------------------- Captured log call --------------------------------------------------------------------------
ERROR    pymmcore_plus.core._mmcore_plus:_mmcore_plus.py:574 Exception occured in MMCorePlus callback 'exposureChanged': <SignalInstance 'exposureChanged' on <pymmcore_plus.core._signals._CMMCoreSignaler object at 0x7fc9a9d3f9d0>>

```

tests/test_main_window.py
```diff
@@ -27,6 +29,72 @@
         )
 
 
+def test_exposure_changing(qtbot: "QtBot", main_window: MainWindow):
+    mmc = main_window._mmc
+    remote = isinstance(mmc, RemoteMMCore)
+
+    # qtbot wait was breaking at times with pysygnal signals
```
ianhi, line 36: In particular with this error:

```
xceptions caught in Qt event loop:
________________________________________________________________________________
  File "/home/ian/mambaforge/envs/pymm/lib/python3.9/site-packages/pytestqt/wait_signal.py", line 58, in _quit_loop_by_timeout
    self._cleanup()
  File "/home/ian/mambaforge/envs/pymm/lib/python3.9/site-packages/pytestqt/wait_signal.py", line 572, in _cleanup
    super()._cleanup()
  File "/home/ian/mambaforge/envs/pymm/lib/python3.9/site-packages/pytestqt/wait_signal.py", line 64, in _cleanup
    self._timeout_message = self._get_timeout_error_message()
  File "/home/ian/mambaforge/envs/pymm/lib/python3.9/site-packages/pytestqt/wait_signal.py", line 344, in _get_timeout_error_message
    error_message = self._get_expected_and_actual_signals_message()
  File "/home/ian/mambaforge/envs/pymm/lib/python3.9/site-packages/pytestqt/wait_signal.py", line 519, in _get_expected_and_actual_signals_message
    self._get_signal_string_representation_for_index(missing_signal_index)
  File "/home/ian/mambaforge/envs/pymm/lib/python3.9/site-packages/pytestqt/wait_signal.py", line 555, in _get_signal_string_representation_for_index
    signal_str_repr = self._signal_names[signal]
KeyError: <SignalInstance 'exposureChanged' on <pymmcore_plus.core._signals._CMMCoreSignaler object at 0x7fc9a9d3f9d0>>
________________________________________________________________________________
-------------------------------------------------------------------------- Captured log call --------------------------------------------------------------------------
ERROR    pymmcore_plus.core._mmcore_plus:_mmcore_plus.py:574 Exception occured in MMCorePlus callback 'exposureChanged': <SignalInstance 'exposureChanged' on <pymmcore_plus.core._signals._CMMCoreSignaler object at 0x7fc9a9d3f9d0>>

```

ianhi:
@tlambert03 bump here

ianhi:
Needs a rebase and also putting off for #111

tlambert03:
@ianhi, can you make a new PR with _just_ the bit that prevents the displayed value from matching the actual value?  then do cache in a different pr?

tlambert03:
@ianhi, is this still something that you want to work on, or should we close?

PR #80, opened 2021-12-01
Remove special casing - now  handled by pymmc+

ianhi:
Sibling of https://github.com/tlambert03/pymmcore-plus/pull/59


ianhi:
evidently not as handled as I thought :/

PR #89, opened 2022-01-28
remove install_mm

tlambert03:
trying to do the same thing as #80 

closes #80


tlambert03:
@ianhi, take a quick peek?

This also:

- uses a process-local CMMCore instance, unless started with `MM_REMOTE`
- adds a note to the readme for users to run `python -m pymmcore_plus.install`

.github/workflows/test.yml
```diff
@@ -35,41 +35,19 @@ jobs:
...
         run: |
           python -m pip install --upgrade pip
           pip install -e .[testing]
-          pip install pytest-cov
-
-      - uses: actions/cache@v2
-        id: cache
-        with:
-          path: micromanager_gui/Micro-Manager-*
-          key: ${{ hashFiles('micromanager_gui/install_mm.py') }}
```
ianhi, line 62: Any reason not to do a monthly//weekly expiring cache like in pymmcore-plus? https://github.com/tlambert03/pymmcore-plus/blob/19cf062150aea77953f2864d7738fb7ccc3c2ec3/.github/workflows/test_and_deploy.yml#L39-L50

README.md
```diff
@@ -30,6 +30,27 @@ You can install `napari-micromanager` via [pip]:
...
+The easiest way to get the micromanager adapters is to use:
+
+```
+python -m pymmcore_plus.install
+```
+
+this will install micromanager to the pymmcore_plus folder in your site-package; use this to see where:
+
+```
+python -c "from pymmcore_plus import _util; print(_util.find_micromanager())"
```
ianhi, line 44: We should maybe expose this publicly on pymmcore-plus?

README.md
```diff
@@ -30,6 +30,27 @@ You can install `napari-micromanager` via [pip]:
...
+```
+
+alternatively, you can direct pymmcore_plus to your own micromanager installation with the `MICROMANAGER_PATH`
+environment variable:
```
ianhi, line 48: I think that it will also pick up the default install location? So see main comment, I think there's some value in also advertising that it works with the current MM setup.

micromanager_gui/main_window.py
```diff
@@ -114,7 +115,19 @@ def __init__(self, viewer: napari.viewer.Viewer, remote=True):
         self.objectives_cfg = None
 
         # create connection to mmcore server or process-local variant
-        self._mmc = RemoteMMCore() if remote else CMMCorePlus()
+        _v = bool(os.getenv("MM_VERBOSE"))
+        self._mmc = RemoteMMCore(verbose=_v) if remote else CMMCorePlus()
```
ianhi, line 119: This is maybe pre-mature? without any changes to threading the default will freeze up the UI. 

Also should document this in the readme.

ianhi (commented):
Few thoughts:

1. caching
This mostly speeds up CI - what's the motivation for removing pip install caching (e.g. napari and other deps)

2. MM install location

I think it would be good also make mention that this will work with the standard MM install location without extra effort - thus making it easier for people to try this out when considering switching over from Micromanager.  The advantage of the way you've done it here I think will become obvious once the device drivers can be installed as their own package and could be local to each environment. But until then making it clear that extra steps aren't necessary seems like a net positive. 

3. Thread local

Need to document the environment options. But mostly think this shold wait until improvements in pymmcore-plus, or threading added here to avoid UI freezes.

README.md
tlambert03: agreed!

micromanager_gui/main_window.py
tlambert03: yeah, can do in another PR.  however, I _think_ most of the things we actually need to be threaded are within pymmcore (not here)

ianhi: My understanding of device drivers is that modern ones (and some of the older ones) spawn threads and return immediately from functions, with the notable exception of `snap_image`

tlambert03:
`test_explorer` is now  failing because of https://github.com/tlambert03/pymmcore-plus/issues/79

micromanager_gui/main_window.py
```diff
@@ -84,7 +85,7 @@ def setup_ui(self):
         uic.loadUi(self.UI_FILE, self)  # load QtDesigner .ui file
 
         # set some defaults
-        self.cfg_LineEdit.setText("demo")
+        self.cfg_LineEdit.setText("MMConfig_demo.cfg")
```
ianhi, line 88: see https://github.com/tlambert03/napari-micromanager/pull/95 for a more complete fix here

Issue #29, opened 2021-07-27
Allow for multiple focus devices


ianhi:
Microscopes can have multiple focus devices (e.g. stage controls + perfect focus). Micromanager handles this by allowing toggling between them:
![image](https://user-images.githubusercontent.com/10111092/127217325-f8e4b587-f1e5-4502-9adb-be825c5f7ad9.png)

napari-micromanger should also allow choosing which device is used.

ianhi:
This is fixed in pymmcore-widgets

PR #90, opened 2022-01-31
Make the gui easier to modify by splitting the main widget in multiple sub-widgets

fdrgsp:
Instead of having a single `.ui` file for the whole `napari-micromanager` gui, now there are multiple sub-widgets in the `_gui_objects` folder.

`.ui` files are used only for the more complex `mm_xyz_stage.ui`, `multid_gui.ui`, and `explore_sample.ui`.

The gui is now created in `_gui.py`.

In this PR I have also to the gui the comboboxes to allow to select multiple stage devices (as suggested in #29) (for now only XY and Z. Offset/Autofocus devices not implemented yet).

Screen Shot 2022-02-21 at 6 27 40 PM


ianhi:
@fdrgsp I'm happy to review this, but if you could do the below two things it would make it much easier to do so:

1. Mark as draft until you think it is done (is it done already?)
2. Do a rebase and squash your commits into a few commits where each commit is one logical change. This would make it much easier to review.
You can see how to squash commits together here: https://thoughtbot.com/blog/git-interactive-rebase-squash-amend-rewriting-history#squash-commits-together

tlambert03:
I think this is definitely a step in the right direction!  I'd be inclined to merge this sooner than later to avoid conflicts with other PRs.  @fdrgsp, can you comment on whether you were able to resist the urge to modify _behavior_ for the most part?  All the file changes naturally make it hard to compare the diffs.  I think this is what @ianhi was referring to.  But, if you tried hard to restrict this _just_ to UI/layout changes, with no functionality changes... and if all the tests are passing (which they appear to be, besides windows), then I think I'd be ok generally getting this in and following up with smaller cleanups afterwards.  Can you think of anything you might have touched that wasn't strictly UI related?

One thing that I think is changing here is layout padding.  So the UI is starting to take up more empty space.  You might want to try `setContentsMargins(0,0,0,0)` on some of the new layout objects to avoid all that dead space.

Sidenote (_not_ to be added to this PR!) I think we should either start letting some of these subwidgets be their own thing,  such that they can be added to the window and popped out independently of the main widget.   Alternatively, we could start using QCollapsible from superqt to save space (as shown below):

https://user-images.githubusercontent.com/1609449/154854854-5cf46307-a478-427c-80cb-651d183e4659.mov

but let's definitely not do that here.

fdrgsp:
@tlambert03 

> Can you think of anything you might have touched that wasn't strictly UI related?

No I don't think I modified any behavior. The only think I have changed a little is in `multid_widget.py` and `explore_sample.py`. 
- In `multid_widget.py` I added the possibility to run the experiment even if there is not a motorized xy and/or z stage (see `_get_state_dict()`).

- For `explore_sample.py`, in `main_window.py` I disabled the explorer tab if there is no xy stage.

> One thing that I think is changing here is layout padding. So the UI is starting to take up more empty space. You might want to try setContentsMargins(0,0,0,0) on some of the new layout objects to avoid all that dead space.

I'll work on that!

ianhi:
> Sidenote (not to be added to this PR!) I think we should either start letting some of these subwidgets be their own thing, such that they can be added to the window and popped out independently of the main widget. Alternatively, we could start using QCollapsible from superqt to save space (as shown below):

wow yes please! yes to both probably, but QCollapsible as a start that can (per above gif) be implemented quickly

fdrgsp:
> One thing that I think is changing here is layout padding. So the UI is starting to take up more empty space. You might want to try setContentsMargins(0,0,0,0) on some of the new layout objects to avoid all that dead space.

@tlambert03 I think it looks better now

tlambert03:
@ianhi, lemme know if you want to peek at this one.  Will likely merge tomorrow (this one is very prone to merge conflicts)

micromanager_gui/main_window.py
```diff
@@ -151,34 +103,38 @@ def __init__(
...
 
         # connect comboBox
-        self.objective_comboBox.currentIndexChanged.connect(self.change_objective)
-        self.bit_comboBox.currentIndexChanged.connect(self.bit_changed)
-        self.bin_comboBox.currentIndexChanged.connect(self.bin_changed)
-        self.snap_channel_comboBox.currentTextChanged.connect(self._channel_changed)
+        self.obj.objective_comboBox.currentIndexChanged.connect(self.change_objective)
+        self.cam.bit_comboBox.currentIndexChanged.connect(self.bit_changed)
+        self.cam.bin_comboBox.currentIndexChanged.connect(self.bin_changed)
+        self.tab.snap_channel_comboBox.currentTextChanged.connect(self._channel_changed)
```
ianhi, line 129: I think it makes more sense to move these signal connections (and respective callback methods) into the sub widgets. It feels a bit awkward to have the main window be responsible for managing the state of the sub widgets.

Also this would probably force us to be better about using the `mmc` object as the real source of truth (sorta the model of MV framework) rather than having that work leak into the main window.

micromanager_gui/main_window.py
```diff
@@ -187,41 +143,166 @@ def __init__(
...
+                data = self._mmc.getLastImage()
+            except (RuntimeError, IndexError):
+                # circular buffer empty
+                return
+        try:
+            preview_layer = self.viewer.layers["preview"]
+            preview_layer.data = data
+        except KeyError:
+            preview_layer = self.viewer.add_image(data, name="preview")
+
```
ianhi, line 237: is there any changes in this whole section? I'm not sure why this is showing up on github, but a lot of this I think is showing up due to how you pulled in new changes (as commits rather than merge commits or rebases)

micromanager_gui/_gui_objects/_mm_configuration_widget.py
```diff
@@ -0,0 +1,51 @@
...
+        self.cfg_groupBox = QtW.QGroupBox()
+        self.cfg_groupBox.setTitle("Micro-Manager Configuration")
+        self.main_layout.addWidget(self.cfg_groupBox, 0, 0)
+        self.setLayout(self.main_layout)
+
+        # define camera_groupBox layout
+        self.cfg_groupBox_layout = QtW.QGridLayout()
+
+        # add to cfg_groupBox layout:
+        self.cfg_LineEdit = QtW.QLineEdit()
```
ianhi, line 32:
```suggestion
        self.cfg_LineEdit = QtW.QLineEdit()
        self.cfg_LineEdit.setPlaceHolderText("MMConfig_demo.cfg")
```

ianhi: this placeholder got lost in the move away from `ui` file.

micromanager_gui/_gui_objects/_camera_widget.py
```diff
@@ -0,0 +1,124 @@
...
+        # bin widget in groupbox
+        self.camera_groupBox_layout.addWidget(self.bin_wdg, 0, 0)
+
+        # bit widget and layout
+        self.bit_wdg = QtW.QWidget()
+        self.bit_layout = QtW.QGridLayout()
+        self.bin_layout.setSpacing(0)
+        self.bit_layout.setContentsMargins(0, 0, 3, 0)
+        # label bit in groupbox r1 c0
+        self.bit_label = QtW.QLabel(text="Bit Depth:")
```
ianhi, line 61: For me at least this is not rendering correctly:
image

ianhi: turns out this was bad before for me. Not a blocker

ianhi (changes requested):
Overall I'm a big fan of this! I think it will it much easier to get a handle on what's happening for any new contributors and also makes it easier to make changes and improvements!

That said I'd suggest holding off a little bit for two reasons:

1. (Simple) fix placeholder text in config lineedit - see comment
2. Moving these around is good time to consider which widgets are responsible for managing what state and where signals should be connected. 


For 2 I'd suggest that each of the subwidgets is responsible for managing the relevant core signals.  And each subwidget should provide a `refresh` method that the main window can call when a new config. 

not 100% on this, but think that now is a good time to make a decision as to where signal connections should be happening.

micromanager_gui/_gui_objects/_camera_widget.py
fdrgsp: As in #107 I think we can remove the widgets for binning and bit depth.

micromanager_gui/_gui_objects/_mm_configuration_widget.py
fdrgsp: I'll put it back in the widget.

micromanager_gui/main_window.py
fdrgsp: I think this is just because I reordered the methods

micromanager_gui/main_window.py
fdrgsp: @ianhi  I think this can be a good idea! However it is maybe better to merge this "new" way to make the gui before doing any further changes...otherwise it might be harder to deal with conflicts. In addition, once this PR is merged, it will be quite easy to merge #107 which can be the new starting point to change widgets signals, etc.... why do you think?

tlambert03:
> That said I'd suggest holding off a little bit for two reasons:
> 2. Moving these around is good time to consider which widgets are responsible for managing what state and where signals should be connected.
> For 2 I'd suggest that each of the subwidgets is responsible for managing the relevant core signals. And each subwidget should provide a `refresh` method that the main window can call when a new config.
> not 100% on this, but think that now is a good time to make a decision as to where signal connections should be happening.


In general, I agree with all these comments.  I definitely agree that in general each subwidget should be more self-contained, and main window should simply compose them all.  Signals should be contained on subwidgets as much as possible, and state should be made clearer (as in #109).  I don't, however, think that this PR is a good time to change all that.  It's not a new problem, but having this PR in would make it easier to reason about all that stuff and fix in smaller followup PRs.  Meanwhile, as long as this stays open, pretty much every other PR is going to have to either wait or cause a conflict that requires manual resolution.

If we generally feel like this is a good direction, i think we clean up the little remaining issues and merge... and then (**!**) lets get back to much smaller PRs that try to solve a single problem.  @fdrgsp, lemme know when you're done making changes and I'll do a final thorough review

fdrgsp:
>  lemme know when you're done making changes and I'll do a final thorough review

@tlambert03 I think you can review it!

tlambert03:
two quick observations you can work on while I continue to look:

1) why does `MainWindow` call `self.create_gui` instead of `MicroManagerWidget` ... when it seems to be entirely contained to `MicroManagerWidget`?  

2) something has been lost with regards to enabling the gui.  On main, if you run `python launch-dev.py`, everything is active and ready to go.  Here, it appears that one _must_ interact with the GUI (and click the load button) to make everything active.  We definitely have to preserve the "programmatic" activation as well.  The gui should be smart enough to know if it's ready to go, without expecting a specific manual action in the GUI

tests/conftest.py
```diff
@@ -22,7 +22,10 @@ def main_window(qtbot, request):
     viewer = Viewer(show=False)
     win = MainWindow(viewer=viewer, remote=request.param == "remote")
     config_path = os.path.dirname(os.path.abspath(__file__)) + "/test_config.cfg"
-    win._mmc.loadSystemConfiguration(config_path)
+    win.cfg.cfg_LineEdit.setText(config_path)
+    win.load_cfg()
+
+    win._set_enabled(True)
```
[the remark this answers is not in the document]
fdrgsp: @tlambert03 

Something like this?

```
...
sig.systemConfigurationLoaded.connect(self._on_system_cfg_loaded)
...
def _on_system_cfg_loaded(self):
        if len(self._mmc.getLoadedDevices()) > 1:
            self._set_enabled(True)
```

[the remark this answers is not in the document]
tlambert03: sure, if that works!

ianhi:
> something has been lost with regards to enabling the gui. On main, if you run python launch-dev.py, everything is active and ready to go. Here, it appears that one must interact with the GUI (and click the load button) to make everything active. We definitely have to preserve the "programmatic" activation as well. The gui should be smart enough to know if it's ready to go, without expecting a specific manual action in the GUI

On main this happens due to these lines:
https://github.com/tlambert03/napari-micromanager/blob/eac0da7c33671ff7210e1f6e9ec5a9446329a457/micromanager_gui/main_window.py#L183-L184

> but having this PR in would make it easier to reason about all that stuff and fix in smaller followup PRs. Meanwhile, as long as this stays open, pretty much every other PR is going to have to either wait or cause a conflict that requires manual resolution.

> If we generally feel like this is a good direction, i think we clean up the little remaining issues and merge... and then (!) 


Yup sounds good! My only worry here is that there are some other small things like the programmatic launching that may have been lost bc we don't have complete tests and the diff is so big. But short of breaking this up into a series of much smaller PRs I don't think there much of a way around that. So modulo fixing the two issues talley noted :+1:  from me.

tlambert03:
just pushed some changes @fdrgsp.  Couple things to point out:

1) i saw a lot of QGridLayout usage when all that was needed was a QVBox/QHBoxLayout.  In general, if you don't _need_ a grid, use one of the simpler layouts instead
2) i noticed a number of attributes being named one thing in the MicroManagerWidget (eg. `self.mm_configuration`) only to be renamed (eg. `self.cfg`) in the MainWindow subclass.  Try not to rename things gratuitously as it makes it harder to get a sense for what a class really _is_ (for example, now there's a mental burden to remember that those two things are actually just two names for the same thing)

If tests pass, I will probably merge... since I think this is heading in the right direction and I'm rather scared of merge conflicts now.  After this merges, I'll merge it into #107  and merge that too.  

_then_.... I'd really like to enforce #111.  I'd like some time to go through all the code, minimize duplication and simplify code, and shore up our tests.  I agree with @ianhi's statement that it's scary to make changes like this without having great test coverage.  so let's get that done before moving forward.

thanks for this big effort @fdrgsp!

PR #93, opened 2022-02-09
use threads for snap

ianhi:
Use a thread when snapping to prevent the UI freezing when using `CMMCorePlus`.

I tried to use the napari generator threading example to only need to spawn one worker, but I couldn't consistently get it to not accidentally snap twice so spawning a new one everytime (unless this is a bad thing?)

I also tried to add the option of psasing and existing core object in to allow launching napari from another script and still having access to the core object (aside from `main_window._mmc` but I think that this is limited by https://github.com/napari/napari/issues/1211. So maybe it would be good for the process local pymmcore-plus to have a mechanism to get an already existing CMMCore object if one exists?

The other issue I'm noticing that probably will need fixing on pymmcore-plus is that now that we've introduced threading you can do bad things like
1. snap an image with a long exposure
2. change some other setting (intentionally or accidentally) that send a command to mmc
That's bad in the sense that it would mess up your image, but also I'm not sure how `MMCore` would even behave there, potentially poorly I imagine.
For example if you mash the `snap` button and then hit the `live` button you get errors like this:
```
    600 @thread_worker(
    601     connect={"finished": lambda: self.update_viewer(self._mmc.getImage())}
    602 )
    603 def snap_worker():
--> 604     self._mmc.snapImage()
        self._mmc = <CMMCorePlus at 0x7fab9e0e2f40>
        self = <micromanager_gui.main_window.MainWindow object at 0x7fabb2ddc3a0>

RuntimeError: This operation can not be executed while sequence acquisition is running.
```

so there may need to be some sort of locking mechanism on pymmcore-plus?

### DIFF
--- a/micromanager_gui/main_window.py
+++ b/micromanager_gui/main_window.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING
 
 import napari
 import numpy as np
+from napari.qt.threading import thread_worker
 from pymmcore_plus import CMMCorePlus, RemoteMMCore
 from pymmcore_plus._util import find_micromanager
 from qtpy import QtWidgets as QtW
@@ -101,7 +102,12 @@ class _MainUI:
 
 
 class MainWindow(QtW.QWidget, _MainUI):
-    def __init__(self, viewer: napari.viewer.Viewer, remote=True):
+    def __init__(
+        self,
+        viewer: napari.viewer.Viewer,
+        remote=False,
+        mmc: CMMCorePlus | RemoteMMCore = None,
+    ):
         super().__init__()
         self.setup_ui()
 
@@ -112,7 +118,10 @@ class MainWindow(QtW.QWidget, _MainUI):
         self.objectives_cfg = None
 
         # create connection to mmcore server or process-local variant
-        self._mmc = RemoteMMCore() if remote else CMMCorePlus()
+        if mmc is not None:
+            self._mmc = mmc
+        else:
+            self._mmc = RemoteMMCore() if remote else CMMCorePlus.instance()
 
         adapter_path = find_micromanager()
         if not adapter_path:
@@ -594,8 +603,15 @@ class MainWindow(QtW.QWidget, _MainUI):
 
     def snap(self):
         self.stop_live()
-        self._mmc.snapImage()
-        self.update_viewer(self._mmc.getImage())
+
+        # snap in a thread so we don't freeze UI when using process local mmc
+        @thread_worker(
+            connect={"finished": lambda: self.update_viewer(self._mmc.getImage())}
+        )
+        def snap_worker():
+            self._mmc.snapImage()
+
+        snap_worker()
 
     def start_live(self):
         self._mmc.startContinuousSequenceAcquisition(self.exp_spinBox.value())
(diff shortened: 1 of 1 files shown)


ianhi:
I had a go at some of the pymmcore-plus changes here: https://github.com/tlambert03/pymmcore-plus/pull/73

ianhi:
@tlambert03 on tests passing I think this is ready

setup.cfg
```diff
@@ -29,8 +29,7 @@ include_package_data = True
 install_requires =
     napari>=0.4.13
     scikit-image
-    tqdm
-    pymmcore-plus >=0.1.8
+    pymmcore-plus >=0.2.0
```
tlambert03, line 32:
```suggestion
    pymmcore-plus >=0.2.1
```

(releasing now, with the fix to setPosition)

ianhi: done

micromanager_gui/main_window.py
```diff
@@ -114,7 +120,10 @@ def __init__(self, viewer: napari.viewer.Viewer, remote=True):
         self.objectives_cfg = None
 
         # create connection to mmcore server or process-local variant
-        self._mmc = RemoteMMCore() if remote else CMMCorePlus()
+        if mmc:
```
tlambert03, line 123:
```suggestion
        if mmc is not None:
```
rarely (though likely not in this case), objects have odd `__bool__` behavior that makes it a bit safer to explicitly check `not None`

tlambert03 (approved):

tlambert03:
can you merge main on this so tests can pass?

micromanager_gui/main_window.py
```diff
@@ -594,8 +603,15 @@ def update_max_min(self, event=None):
 
     def snap(self):
         self.stop_live()
-        self._mmc.snapImage()
-        self.update_viewer(self._mmc.getImage())
+
+        # snap in a thread so we don't freeze UI when using process local mmc
+        @thread_worker(
+            connect={"finished": lambda: self.update_viewer(self._mmc.getImage())}
+        )
```
ianhi, line 610: what if instead of doing an actoin on "finished" we just connect to the new `imageSnapped` signal?

pro:

Allows synchronizing with a script (e.g. notebook) so when you snap there it also shows up in napari

cons:
maybe complicated when in an mda?
I suppose this draws us further away from other backends, although that is already probably a decent bit away.

micromanager_gui/main_window.py
```diff
@@ -594,8 +603,15 @@ def update_max_min(self, event=None):
...
+        def snap_worker():
+            self._mmc.snapImage()
+
+        snap_worker()
```
tlambert03, line 614:
```suggestion
        # snap in a thread so we don't freeze UI when using process local mmc
        create_worker(
            self._mmc.snapImage,
            _connect={"finished": lambda: self.update_viewer(self._mmc.getImage())},
            _start_thread=True,
        )
```

can use a lower-level function here (lower than the thread_worker decorator)  use `from superqt.utils import create_worker`

micromanager_gui/main_window.py
tlambert03: yeah, slightly wary of depending on the core-plus API at the moment.  Let's start with this.

tlambert03 (commented):
looks good, lets use `superqt.utils.create_worker` directly, then good to merge

ianhi:
everything passing!

PR #96, opened 2022-02-11
Add `main_window_events` + Cycle through channels when adding to MDA

ianhi:
1. Make events for main window
Handy for letting things like the mda widget keep track of overall state

2. cycle through channels when adding
previously each channel would be added with whatever the first channel was. Now we do a naive cycling through them. I explored keeping track of what had been added and removed but seems to be more trouble than it was worth.

Peek 2022-02-11 15-42

### DIFF
--- a/micromanager_gui/_signals.py
+++ b/micromanager_gui/_signals.py
@@ -1,0 +1,9 @@
+from qtpy.QtCore import QObject, Signal
+
+__all__ = [
+    "main_window_events",
+]
+
+
+class main_window_events(QObject):
+    availableChannelsChanged = Signal(tuple)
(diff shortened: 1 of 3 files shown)


micromanager_gui/_signals.py
```diff
@@ -0,0 +1,9 @@
+from qtpy.QtCore import QObject, Signal
+
+__all__ = [
+    "main_window_events",
+]
+
+
+class main_window_events(QObject):
+    availableChannelsChanged = Signal(tuple)
```
ianhi, line 9: Introduction of an events object for the widget is probably the biggest change here

tlambert03:
@ianhi  ... do you still want this?  My gut feeling is that this should be a `valueChanged` signal now on a channel table in  widgets

PR #99, opened 2022-02-17
Disable xyz stage if there is not a xyz device

### LINKED ISSUE #98: [Bug] no protection against moving when there are no XYZ devices
## Problem

The XY (or Z) buttons have no protection against being called even if there is no valid device. So if you click them then they will throw this error:
```
RuntimeError                              Traceback (most recent call last)
File ~/Documents/oss/micro/napari-micromanager/micromanager_gui/main_window.py:497, in MainWindow.stage_y_up(self=<micromanager_gui.main_window.MainWindow object>)
    496 def stage_y_up(self):
--> 497     self._mmc.setRelativeXYPosition(
        self._mmc = <CMMCorePlus at 0x7fb4f961cf40>
        self = <micromanager_gui.main_window.MainWindow object at 0x7fb4f9fa8f70>
        self.xy_step_size_SpinBox = <PyQt5.QtWidgets.QSpinBox object at 0x7fb4f9595280>
    498         0.0,
    499         float(self.xy_step_size_SpinBox.value()),
    500     )
    501     if self.snap_on_click_xy_checkBox.isChecked():
    502         self.snap()

RuntimeError: No device with label ""

```

## Solution

disable those buttons if there is no device available to be moved.

fdrgsp:
as @ianhi pointed out in #98, this PR disable the XY and/or Z stage controls if there are no xy and/or z devices.

closes #98


ianhi:
@fdrgsp This is a good first pass! But I think we can be a bit more specific about the enabling and disabling. For example in the case where we have a cube switcher, camera and Z stage, but no XY stage we should only disable the `XY` buttons rather than the full gui. Though I think that will take a decent bit more work to implement :(

ianhi:
> @ianhi I thought I was doing this in the _refresh_positions() method. Is this what you meant?

Wow I just missed then when skimming over, maybe i didn't sleep enough.... sorry about that! I'll have a go at trying this branch out later to properly test

fdrgsp:
> > @ianhi I thought I was doing this in the _refresh_positions() method. Is this what you meant?
> 
> Wow I just missed then when skimming over, maybe i didn't sleep enough.... sorry about that! I'll have a go at trying this branch out later to properly test

Great! :)

micromanager_gui/main_window.py
```diff
@@ -272,6 +278,8 @@ def _on_mda_finished(self, sequence: useq.MDASequence):
     def browse_cfg(self):
         self._mmc.unloadAllDevices()  # unload all devicies
         print(f"Loaded Devices: {self._mmc.getLoadedDevices()}")
+        # disable gui
+        self._set_enabled(False)
```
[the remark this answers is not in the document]
tlambert03: seems good, agree @fdrgsp.  can merge after that

fdrgsp:
@ianhi  @tlambert03 Sure! But then I think it is better to move everything in `load_cgf` like in this last commit. In this way if you cancel everything stays as it was. What do you think?

micromanager_gui/main_window.py
```diff
@@ -270,8 +276,19 @@ def _on_mda_finished(self, sequence: useq.MDASequence):
...
+        current_cgf = self.cfg_LineEdit.text()
+
+        file_dir = QtW.QFileDialog.getOpenFileName(self, "", "", "cfg(*.cfg)")
+        self.cfg_LineEdit.setText(str(file_dir[0]))
+        self.max_min_val_label.setText("None")
+        self.load_cfg_Button.setEnabled(True)
+
+        if not self.cfg_LineEdit.text():
+            self.cfg_LineEdit.setText(current_cgf)
+            self.load_cfg_Button.setEnabled(False)
```
ianhi, line 289: I think this can be simplified like so:
```suggestion
        file_dir = QtW.QFileDialog.getOpenFileName(self, "", "", "cfg(*.cfg)")
        if file_dir:
            self.cfg_LineEdit.setText(str(file_dir[0]))
            self.max_min_val_label.setText("None")
            self.load_cfg_Button.setEnabled(True)
```

eliminating the `else` branch because if nothing is returned there may still be text in the lineedit that we want to be able to load.

fdrgsp:
> Ah indeed. See comment, this is currently a little bit broken. If you cancel the dialog then you are no longer allowed to type in the lineedit:

@ianhi I believe this is what I wanted to achieve...the `lineEdit` is set to `ReadOnly` so that you cannot modify the text manually but only with the "browse" button (`...`). 
If you click on "cancel", the `lineEdit` will display the current `.cfg` path. After clicking on "cancel" you can either continue using micromanager with the current config loaded or load a new one by clicking on the "browse" button. Does it make sense? Or do you prefer to let the user modify the `lineEdit` manually?

> eliminating the else branch because if nothing is returned there may still be text in the lineedit that we want to be able to load.

If I don't use it, after clicking on "cancel" the `lineEdit` is empty. Not sure why...That's why I've added:
```
if not self.cfg_LineEdit.text():
      self.cfg_LineEdit.setText(current_cgf)
      self.load_cfg_Button.setEnabled(False)
```
Does this happen to you?

micromanager_gui/main_window.py
ianhi: oops should really be:
```suggestion
        (filename, selectedFilter) = QtW.QFileDialog.getOpenFileName(self, "", "", "cfg(*.cfg)")
        if filename:
            self.cfg_LineEdit.setText(filename)
            self.max_min_val_label.setText("None")
            self.load_cfg_Button.setEnabled(True)
```
in order to not accidently blank out the dialog

ianhi:
>  Or do you prefer to let the user modify the lineEdit manually?

I think having the option to do this is a good thing, they can still use the file selection dialog if they want, but then this makes it easy to do something like, add and `_2` to the name 


> If I don't use it, after clicking on "cancel" the lineEdit is empty. Not sure why...That's why I've added:
> Does this happen to you?

looks like there was a bug in my suggestion. With the new suggestion https://github.com/tlambert03/napari-micromanager/pull/99#discussion_r810702725 it won't blank it out.

ianhi:
>  > Or do you prefer to let the user modify the lineEdit manually?
> I think having the option to do this is a good thing, they can still use the file selection dialog if they want, but then this makes it easy to do something like, add and _2 to the name

But I think this has always been the behavior so we can change that in separate PR, rather than let this one continue to creep in scope.

ianhi:
Final thought for @fdrgsp I think you need to rebase this on main because https://github.com/tlambert03/napari-micromanager/pull/95 was merged. I would do the following after having checked out this branch

```
git fetch upstream
git rebase upstream/main
```

then follow the rebase instructions (fixing the conflicts, then `git add` then `git rebase --continue`)

fdrgsp:
@ianhi 

> But I think this has always been the behavior so we can change that in separate PR, rather than let this one continue to creep in scope.

That's ok, I just switch the readonly property to false.


> Final thought for @fdrgsp I think you need to rebase this on main because #95 was merged. I would do the following after having checked out this branch
> 
> ```
> git fetch upstream
> git rebase upstream/main
> ```
> 
> then follow the rebase instructions (fixing the conflicts, then `git add` then `git rebase --continue`)

I'm not sure I've done what you suggested...I've never use rebase before...

ianhi:
> I'm not sure I've done what you suggested...I've never use rebase before...

I think you merged in the main branch which accomplishes pretty much the same thing. That said I think rebasing is great and particularly like using it in order to combine commits (or I use `git commit --amend`). But it's definitely confusing the first few times you have a go at it. If you ever want to try rebasing insteead of merging I'd be happy to help walk you through it.

tlambert03:
> I think you merged in the main branch which accomplishes pretty much the same thing.

yeah, this is what I showed @fdrgsp.  I generally use merge instead of rebase, since you get it done in one step, and i have squash commits during merge on all my repos anyway (since the vast majority of contributors aren't going to be paying close attention to the quality of each commit).  Rebasing is definitely worth learning!  but I can take the blame for @fdrgsp using merge as I rarely rebase myself :)

tlambert03 (approved):

tlambert03:
quick tip @fdrgsp... if you open a PR to address an issue (such as this one does for #98), then put `closes #98` or `fixes #98` in the original post.  It will then auto-close the linked issue when this merges so we don't have to remember to go do it ourselves.  See:
https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

tlambert03:
please resolve this one too

micromanager_gui/main_window.py
```diff
@@ -213,7 +213,6 @@ def load_cfg(self):
         cfg = self.cfg_wdg.cfg_LineEdit.text()
         if cfg == "":
             cfg = "MMConfig_demo.cfg"
-            self.cfg_wdg.cfg_LineEdit.setText(cfg)
```
ianhi, line 216: Why remove this?

fdrgsp: @ianhi  Sorry, my mistake when fixing conflicts! I'll put it back

tlambert03: actually... do we really want to set it?  I see that we need to fallback to the string `"MMConfig_demo.cfg"` if nothing was provided, but why actually set the text in the LineEdit?  I kinda like that it's just placeholder text until a user manually enters something...

micromanager_gui/main_window.py
```diff
@@ -213,8 +213,8 @@ def load_cfg(self):
...
         self._mmc.loadSystemConfiguration(cfg)
+        self.cfg_wdg.cfg_LineEdit.setText(cfg)
```
tlambert03, line 218:
```suggestion
        cfg = self.cfg_wdg.cfg_LineEdit.text() or "MMConfig_demo.cfg"
        self._mmc.loadSystemConfiguration(cfg)
```
maybe just this?

tlambert03: (I pushed this and will resolve this)

PR #103, opened 2022-02-21
Use instance() when using local pymmcore+

ianhi:
Also remove tqdm dependency as it is no longer used.


ianhi:
Should be merged in concert with https://github.com/tlambert03/napari-micromanager/pull/93

ianhi:
wait I should just modify that PR nevermind.

PR #105, opened 2022-02-25
[WIP] Add autofocus device control

fdrgsp:
This PR (to be eventually merged after #90 ) adds the autofocus control in the gui. At the moment is setup only for the Nikon Perfect Focus System but we can easily add any autofocus device by specifying `offset_device` and `autofocus_device` in `_autofocus.py`.

I will test is on the microscope soon!

Screen Shot 2022-02-25 at 8 20 15 AM


tlambert03:
@fdrgsp  ... same here... too difficult to merge.  please reduce the diff down to the bare minimum, perhaps branch off of main again and force push

tlambert03:
ok with #107 in, go ahead and resolve this one too.

micromanager_gui/main_window.py
```diff
@@ -92,8 +94,10 @@ def __init__(
         sig.systemConfigurationLoaded.connect(self._on_system_cfg_loaded)
         sig.XYStagePositionChanged.connect(self._on_xy_stage_position_changed)
         sig.stagePositionChanged.connect(self._on_stage_position_changed)
+        sig.channelGroupChanged.connect(self._refresh_channel_list)
         sig.exposureChanged.connect(self._on_exp_change)
         sig.frameReady.connect(self._on_mda_frame)
+        sig.propertyChanged.connect(self._on_offset_status_changed)
```
tlambert03, line 100: this connection is too broad.  This will perform the logic in `_on_offset_status_changed` whenever literally _any_ property in mmcore changes... which will happen all the time.  Create an intermediate callback that checks the name of the property being changed, and does nothing if it doesn't need to:

```python
    sig.propertyChanged.connect(self._on_prop_changed)

def _on_prop_changed(self, dev_name: str, prop_name: str, new_value: str):
    if devname_or_propname_indicates_perfect_focus:
        self._on_offset_status_changed()
```

micromanager_gui/_autofocus.py
```diff
@@ -0,0 +1,47 @@
...
+    from pymmcore_plus import CMMCorePlus, RemoteMMCore
+
+
+class AutofocusDevice:
+    def __init__(self, mmcore: CMMCorePlus | RemoteMMCore = None):
+        super().__init__()
+        self._mmc = mmcore
+
+    @classmethod
+    def set(self, key, mmcore):
```
tlambert03, line 15:
```suggestion
    def create(cls, key, mmcore) -> AutofocusDevice:
```

it's convention to use `cls` instead of `self` in classmethods.  Change name to `create` (we're not really setting anything, we're creating a new object).  then update the usage in main_window

micromanager_gui/_autofocus.py
```diff
@@ -0,0 +1,47 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from pymmcore_plus import CMMCorePlus, RemoteMMCore
+
+
+class AutofocusDevice:
+    def __init__(self, mmcore: CMMCorePlus | RemoteMMCore = None):
```
tlambert03, line 10:
```suggestion
    def __init__(self, mmcore: CMMCorePlus | RemoteMMCore):
```

`None` would not ever work right?

ianhi:
I think this may be a good place pause for a second and think about how we want this to be designed. Because autofocus in MMCore can be a complicated and there may be some gotchas. e.g. how does this interact with the MDA panel. Also we should thing about how to signal what the `focusDevice` is. On my setup I have it so the default focus device is the PFS-Offset, which means that the buttons for `Z` actually control that. Some really good description here: https://github.com/micro-manager/mmCoreAndDevices/issues/156


I don't have a specific objection to anything here and suspect that this is an improvement over the current situation, but just want to point out that this is a complex topic that deserves some careful UI design. I would also suggest not strictly replicating what micro-manager does because I've never found that to make much sense.

micromanager_gui/_autofocus.py
tlambert03: (you could also name it `for` if you prefer... as in `AutofocusDevice.for('some_name', mmcore)`

micromanager_gui/_autofocus.py
```diff
@@ -0,0 +1,47 @@
...
+
+class NikonPFS(AutofocusDevice):
+    """
+    for Nikon PFS:
+    - mmcore.getAutoFocusDevice() -> "TIPFStatus"
+    - offset_device to set/get the PFS offset -> "TIPFSOffset"
+    "TIPFSOffset" is listed under StageDevice (z movement).
+    Can be used with mmc.setProperty("TIPFSOffset", "Position", offset)
+    or mmc.getProperty("TIPFSOffset", "Position")
+    """
```
tlambert03, line 44: I'd like to start insisting on proper [PEP 257](https://www.python.org/dev/peps/pep-0257/#specification) docstrings.  Here that would look something like:

```suggestion
class NikonPFS(AutofocusDevice):
    """Nikon Perfect Focus System (PFS) autofocus device.

    To be used when `mmcore.getAutoFocusDevice()` returns `"TIPFStatus"`.
    """
```

(The remainder of the docstring belongs more in the generic AutofocusDevice)

micromanager_gui/multid_widget.py
```diff
@@ -273,6 +275,17 @@ def add_position(self):
                 item.setTextAlignment(int(Qt.AlignHCenter | Qt.AlignVCenter))
                 self.stage_tableWidget.setItem(idx, c, item)
 
+            if self._mmc.getAutoFocusDevice() and self._mmc.isContinuousFocusEnabled():
+                autofocus_dev = AutofocusDevice.set(
+                    self._mmc.getAutoFocusDevice(), self._mmc
+                )
+                autofocus_dev.get_position()
```
tlambert03, line 282: is this doing anything?

fdrgsp:
@ianhi

> I think this may be a good place pause for a second and think about how we want this to be designed...
> On my setup I have it so the default focus device is the PFS-Offset, which means that the buttons for Z actually control that

I designed this based on the way I use the Nikon PFS (a `focus-maintenance` system) and it might not work other focusing systems (and I also was not thinking about any "software-based autofocus"). Since it works with refractive index difference, it cannot be set as the default. And for no particular reason (other than how it looked) I decided to make 2 different widgets, one for z drive and one for offset drive (it could have been only one, and the combobox could have been used to select which one to use). We can definitely discuss on how to actually make it better/much more usable!!!

> how does this interact with the MDA panel.

I was thinking that in the MDA we only need to create the `Sequence` (to be implemented in `useq-schema`) either with Z focus and/or`Z offset` (if exists/engaged) or without any z coordinates (if there is no focusing/offset device) and then `run_mda` in `pymmcore` will use this info to set the z position. What do you think?

> Also we should thing about how to signal what the focusDevice is

I was thinking that this depends on which one the user select before adding/storing positions. In this pr, if the PFS is on end engaged, the z focusing device widget is disabled and vice versa.

> I would also suggest not strictly replicating what micro-manager does because I've never found that to make much sense.

I actually never use "autofocus" systems with micromanager before, what would you change?

micromanager_gui/multid_widget.py
fdrgsp: @tlambert03  This is only adding the `z offset` to the position table if the z offset is engaged. Nothing in particular at the moment since there is no offset info in `useq-schema`/`pymmcore-plus` `run_mda`.

tlambert03: but `autofocus_dev.get_position()` gets the position and doesn't assign it to anything?

fdrgsp: Oh I see, you are right, line 282 is not used!

PR #109, opened 2022-02-25
add set_state method to multid-widget

tlambert03:
This adds one new public method to the multid widget, which allows you to set the state from an MDASequence object, dict, or yaml file

Also added to `launch-dev` for ease of quickly testing an MDA


micromanager_gui/multid_widget.py
```diff
@@ -294,6 +289,94 @@ def set_multi_d_acq_dir(self):
         self.dir_lineEdit.setText(self.save_dir)
         self.parent_path = Path(self.save_dir)
 
+    def set_state(self, state: dict | MDASequence | str | Path) -> None:
```
ianhi, line 292: Similarly would be good to make `get_state` public

tlambert03: Should it return a dict or an MDASequence?

ianhi: default to sequence with optional arg to make it as a dict?

tlambert03: I'll do sequence.  I dislike parameters that change return types.   If someone wants a dict they can always do `get_state().dict()`

tlambert03: made it public, added round-trip test

PR #114, opened 2022-02-28
Ensure `_on_mda_frame` is called in main thread

### LINKED ISSUE #113: Replicate Qt signal thread behavior
### Description
psygnal does not work across threads when used in a pyqt application. (working towards a small example but I know that this is true from pymmcore-plus + napari). 

To fix this I suspect that it will be necessary to implement the different types of connections described here:

https://doc.qt.io/qt-5/threads-qobject.html#signals-and-slots-across-threads

@tlambert03 you also wrote down some thoughts on this here: https://github.com/tlambert03/pymmcore-plus/issues/17#issuecomment-880725024

tlambert03:
Fixes #113 

This one decorator (a brilliant contribution to superqt from @czaki) is all that was needed to fix the errors that happen when the `CMMCore.events` object is backed by `psygnal.Signal` instead of `QtCore.Signal`.

Some notes on how I debugged this:
1. I ran the script provided in #113 using `NAPARI_DEBUG_EVENTS=1 python script.py`
2. that showed that it was consistently the `viewer.add_image()` call in `_on_mda_frame` that caused the crash:
   ```py
    LayerList.events.inserting(index=0)
    was triggered by Viewer.add_image(), via:
    ".../site-packages/napari/utils/events/containers/_evented_list.py", line 177, in LayerList.insert
    ".../site-packages/napari/utils/events/containers/_selectable_list.py", line 65, in LayerList.insert
    ".../site-packages/napari/components/layerlist.py", line 90, in LayerList.insert
    ".../python/_collections_abc.py", line 1073, in LayerList.append
    ".../site-packages/napari/components/viewer_model.py", line 732, in Viewer.add_image
    "/Users/talley/Dropbox (HMS)/Python/forks/napari-micromanager/micromanager_gui/main_window.py", line 348, in MainWindow._on_mda_frame
    "/Users/talley/Dropbox (HMS)/Python/pymmcore-plus/pymmcore_plus/core/_mmcore_plus.py", line 533, in CMMCorePlus._mda
    ".../python/threading.py", line 910, in Thread.run
   ```
4. The error (`WARNING: QObject::connect: Cannot queue arguments of type 'QVector<int>'`) suggests that Qt is trying to communicate across threads.  However, by this point (when we're trying to manipulate the guy) we need to be back in the main thread.
5. Enter @Czaki's magic decorator, that ensures our callback is called in the QApplication's thread.

thanks @Czaki!

### DIFF
--- a/micromanager_gui/main_window.py
+++ b/micromanager_gui/main_window.py
@@ -11,7 +11,7 @@ from pymmcore_plus._util import find_micromanager
 from qtpy import QtWidgets as QtW
 from qtpy.QtCore import Qt, QTimer
 from qtpy.QtGui import QColor, QIcon
-from superqt.utils import create_worker
+from superqt.utils import create_worker, ensure_main_thread
 
 from ._camera_roi import CameraROI
 from ._gui_objects._mm_widget import MicroManagerWidget
@@ -308,6 +308,7 @@ class MainWindow(MicroManagerWidget):
         """ "create temp folder and block gui when mda starts."""
         self._set_enabled(False)
 
+    @ensure_main_thread
     def _on_mda_frame(self, image: np.ndarray, event: useq.MDAEvent):
         meta = self.mda.SEQUENCE_META.get(event.sequence) or SequenceMeta()
 


ianhi:
oh neat! 

Does this mean we can remove the auto signaller selection in pymmcore+? If yes then we should pair that with an example of using this decorator. 

Also, depending on your goals for psygnal it may be worth keeping the issue around, because until it matches the threading behavior it can't be a drop in replacement - though that's only important if that's a goal of the project.

tlambert03:
> Does this mean we can remove the auto signaller selection in pymmcore+? 

Technically... but don't mind that bit in there.  It a Qt event loop, I think it's generally nice to use Qt signals.  And since that code requires no imports (just checks to see if we've already got a QtApp running), it's pretty innocuous.

> If yes then we should pair that with an example of using this decorator.

I kind of agree, but the general lesson of "don't update the gui from a non-gui thread) is bigger than psygnal for sure.  I'm ok with psygnal not yet being the place one is supposed to learn that lesson.  superqt has documentation on this decorator, and that kinda feels like where it belongs.

> Also, depending on your goals for psygnal it may be worth keeping the issue around, because until it matches the threading behavior it can't be a drop in replacement - though that's only important if that's a goal of the project.

yeah, I do agree with this... but I also feel like there was something a bit more magical to this.  Even when we use QThreads and Qt Signals and slots (ie, with the `@thread_worker` decorator in napari) we _still_ frequently have issues if someone were to try to manipulate the gui from a thread.  So I'm not sure I see this as something as simple as "it will always work if you use Qt signals, but almost never work if you use psygnal".  It will generally _not_ work in either case unless you are careful.   I don't like to keep issues open unless they are actionable, so I'd prefer to close it until it rears its head in a concrete way again

Czaki:
> Also, depending on your goals for psygnal it may be worth keeping the issue around, because until it matches the threading behavior it can't be a drop in replacement - though that's only important if that's a goal of the project.

for implement something like this for psygnal there is a need to heave clear concepts of "main thread" and "object thread". The Qt ship one concept there main thread his this one in which QApplication was created and object storage information about its thread and has interface to move object form one thread to another. 

I'm worry that try to add this to psygnal wil change it from signal implementation to much more complex thing.

PR #115, opened 2022-03-01
Establish global core singleton pattern, and do camera widget cleanup

tlambert03:
This is the first of a number of PRs that will aim to pull logic off of MainWindow, and create more independent widgets that serve as views onto small parts of the mmcore model.   While this started as a camera widget cleanup, this one is a bit bigger than I hope the following ones will be, because it also establishes a new global "core singleton" pattern, and a new module `_core.py` that will have utility functions that just interact with core. 

This PR 

1) adds `_core.get_core_singleton`, which will create/retrieve the global core singleton object being used for the session.  The first call to this function determines whether we're running a remote core or a local core (or, you can set `_core._SESSION_CORE = None` to start over, as I do in the tests.
2) establishes a `_core.CoreState` dataclass that serves as a state object for things that pymmcore's CMMCore doesn't really have (like the current objective device.  This object may well go away in the future, but for now, it lets us avoid adding little attributes to the MMCore object itself, while also keeping this core-related state off of MainWindow.
3)  It cleans up the Camera widget (and that module provides a pattern of how future cleanups will go):
   - attributes that don't need to be stored as `self.___` on the widget are removed
   - unnecessary layout objects are removed (e.g. here, there were 3 HBox layouts, but there only needs to be one)
   - removes hardcoded min/max width settings (this rarely work out in the long run, and [size policies](https://doc.qt.io/qt-5/qsizepolicy.html) should be preferred).
   - Pulls off as much camera related logic from MainWindow as possible (I hope to pull off the CameraROI widget as well in the future.  but stopped short of doing that here.
   - `MainWindow._update_pixel_size` becomes `_core.update_pixel_size`


micromanager_gui/_core.py
```diff
@@ -0,0 +1,99 @@
...
+
+    mmc = mmc or get_core_singleton()
+    # if pixel size is already set, and we're not providing a new value, return.
+    if current_px_size_cfg := mmc.getCurrentPixelSizeConfig():
+        if not pixel_size:
+            return
+        mmc.deletePixelSizeConfig(current_px_size_cfg)
+
+    # create and store a new pixel size config for the current objective.
+    obj_dev_label = obj_dev_label or CoreState.instance().objective_device
```
ianhi, line 86: any reason not be consistent with other files and use `STATE`?

ianhi (approved):
Haven't tried it out but everything looks great!

PR #119, opened 2022-03-01
Cleanup objectives widget

tlambert03:
This extracts all the objective-specific logic from the main window and puts it onto the objective widget.

will leave as draft until #118 merges as there is some overlapping stuff there.

Also intend to re-examine the objective setting logic at that point


tlambert03:
logic improved quite a bit.  tests currently failing with:
```py
        mmc.deleteConfigGroup("Objective")
        main_window.obj_wdg._refresh_objective_choices()
>       assert main_window.obj_wdg.combo.currentText() == "Nikon 10X S Fluor"
E       AssertionError: assert 'Startup' == 'Nikon 10X S Fluor'
E         - Nikon 10X S Fluor
E         + Startup
```

i.e. after we delete the `Objective` config group, it erronously thinks the `System` config group is the one with the objectives.  Solution is to just not really look for config groups, but rather just use `getStateLabels` on the putative objective device

micromanager_gui/_core.py
```diff
@@ -118,26 +118,40 @@ def update_pixel_size(
...
+    # in the micromanager docs, `Objective` is used both as the device label and the
+    # configuration group name. So, it's worth checking that first.
+    for cfg_group in sorted(groups, key=lambda x: x != label):
+        for preset in mmc.getAvailableConfigs(cfg_group):
+            for device_label, _, _ in mmc.getConfigData(cfg_group, preset):
+                if device_label == label:
+                    results.append(cfg_group)
+                    _r += 1
+                    break
+            # only need to check the first preset, since preset share the same devices
```
ianhi, line 152: sadly this is not true. It's probably true that the mircomanager gui limits you to this, but if you edit the text of the file then you can use different devices in different presets and everything will still work. You probably shouldn't do that, but I also know that some people have done that

ianhi: I agree that that's likely safer. Though it will almost certainly break for someone out there who switches multiple things in order to switch their magnification. And it may surprise users when their config group isn't being used for that.

ianhi: Can we add docs here as well. I think I'd feel totally fine with that as long as it was well documented in an accessible way

tlambert03:
This is about ready.  I was able to remove the `_core.STATE` thing altogether... but probably broke a little bit of functionality in the process. Now that I have a better idea of what the ultimate goal is (which I believe is to make sure that `mmcore.getPixelSizeUm()` is just always up to date for the explorer widget, I think we should step back and redesign that bit.  I'll open a new issue to get some feedback from @fdrgsp.

So, once this merges, I do expect tiling acquisition to be broken briefly, and that should be the next thing to fix... probably along the same lines of some of the discussion going on over in #134

tests/test_cam_objective.py
```diff
@@ -17,17 +21,21 @@ def test_crop_camera(main_window: MainWindow):
...
+    with pytest.raises(ValueError):
+        obj_wdg._combo.setCurrentText("10asdfdsX")
 
-    main_window.obj_wdg.objective_comboBox.setCurrentText("10X")
-    assert main_window.obj_wdg.objective_comboBox.currentText() == "10X"
-    assert mmc.getCurrentPixelSizeConfig() == "Res10x"
-    assert main_window.cam_wdg.px_size_spinbox.value() == 1.0
+    assert global_mmcore.getCurrentPixelSizeConfig() == "Res10x"
 
-    main_window.cam_wdg.px_size_spinbox.setValue(6.5)
```
tlambert03, line 28: note, removal of this line is temporarily removing the custom pixelSizeConfig functionality

micromanager_gui/main_window.py
```diff
@@ -383,103 +380,6 @@ def _on_config_set(self, groupName: str, configName: str):
...
-            self._mmc.setProperty(
-                _core.STATE.objective_device,
-                "Label",
-                self.obj_wdg.objective_comboBox.currentText(),
-            )
-
-        self._mmc.waitForDevice(_core.STATE.objective_device)
-        self._mmc.setPosition(zdev, currentZ)
-        self._mmc.waitForDevice(zdev)
-        self.cam_wdg._update_pixel_size()
```
tlambert03, line 481: Here too: note that the new widget doesn't yet update pixel size when changing the objective... That needs some further consideration

PR #122, opened 2022-03-02
feat: add group and preset table widget

fdrgsp:
merge after #164 

This PR adds to the gui a table widget (`_group_preset_table_widget.py`) that contains the group and presets retrieved from the loaded system configuration.

I also added 3 new groups to `test_config.cfg` for tests (`_combobox_no_preset_test`, `_lineedit_test`, `_slider_test`).
 
Screen Shot 2022-03-02 at 2 34 41 PM


micromanager_gui/_gui_objects/_group_preset_table_widget.py
```diff
@@ -0,0 +1,120 @@
...
+            max=float(self._mmc.getPropertyUpperLimit(device, property)),
+            label=str(property),
+            name=f"{presets[0]}",
+            annotation=[device, property, value],
+        )
+
+    def _set_widget(self, group, presets) -> Widget:
+        wdg = None
+
+        device, property, value, count = self._get_cfg_data(group, presets[0])
```
tlambert03, line 75: need to catch `if not presets`

micromanager_gui/_gui_objects/_group_preset_table_widget.py
```diff
@@ -0,0 +1,120 @@
...
+
+        if len(presets) > 1:
+            wdg = ComboBox(choices=presets, name=f"{presets}", annotation=[])
```
tlambert03, line 78: that's not really what annotation is for... it should be a type annotation  (but can generally be omitted).  Are you using it to store info or something?

tlambert03 (commented):
A great start!  I think most of the remaining work is on cleaning up _set_widget.  Here's what I would recommend.  Make the entirety of the table code something like this:
```python
class MMGroupPresetTableWidget(QtW.QWidget):
    def __init__(self):
        super().__init__()

        self._mmc = _core.get_core_singleton()
        self._mmc.events.systemConfigurationLoaded.connect(self._populate_table)

        self.table_wdg = MainTable()
        self.table_wdg.column_headers = ("Groups", "Presets")
        self.table_wdg.show()
        self.setLayout(QVBoxLayout())
        self.layout().addWidget(self.table_wdg.native)

    def _populate_table(self):
        self.table_wdg.clear()
        if groups := self._mmc.getAvailableConfigGroups():
            self.table_wdg.value = {
                "data": [[group, make_group_widget(group)] for group in groups],
                "index": [],
                "columns": ["Groups", "Presets"],
            }
```

and then work on creating a new function `make_group_widget` that just takes the config group name (one can get presets from that), that is mostly based on your `_set_widget` code.  It still needs some work though.  It was a little bit hard to parse the intentions of each conditional in there, so a general docstring for the whole function would be greatly appreciated.  It should state the heuristics it uses (such as only looking at the first preset in the group), and briefly describe what happens if there's a single device in the preset, multiple devices, or no device.  Lastly, the `_on_change` function itself shouldn't be making all of those `isinstance()` checks: we want callbacks to generally be fast, and you already _know_ what type of widget it is... so do something like:

```python
if isinstance(wdg, ...):
    def _on_change(value: Any): ...
elif isinstance(wdg, ...):
    def _on_change(value: Any): ...

wdg.changed.connect(_on_change)
```

micromanager_gui/_gui_objects/_group_preset_table_widget.py
```diff
@@ -0,0 +1,120 @@
...
+            if count == 1 and self._mmc.getAllowedPropertyValues(device, property):
+                # if is a combobox without user-defined" presets"
+                # but with devices defined presets (e.g. "Binning"->[1,2,4,8])
+                prs = self._mmc.getAllowedPropertyValues(device, property)
+                wdg = ComboBox(
+                    choices=prs, name=f"{presets}", annotation=[device, property]
+                )
+
+            elif count == 1 and self._mmc.getPropertyLowerLimit(device, property):
+                wdg = self._set_float_slider_wdg(presets, device, property, value)
```
tlambert03, line 91: This section here would be great to break out into a `PropertyWidget` akin to the StateDeviceWidget in #120.
This might actually make for a good PR before this PR.  It should just take the name of the device and the name of the property, and build an appropriate widget.  Use things like `core.getPropertyType`, `core.hasPropertyLimits`, and `core.getAllowedPropertyValues` to determine how to build the single widget.  Then connect it (in both directions) to `core.events.propertyChanged` as well as to the widget change event

micromanager_gui/_gui_objects/_group_preset_table_widget.py
```diff
@@ -0,0 +1,120 @@
...
+
+        self.table_wdg = MainTable()
+        self.table_wdg.column_headers = ("Groups", "Presets")
+        self.table_wdg.show()
+        self.setLayout(QVBoxLayout())
+        self.setContentsMargins(0, 0, 0, 0)
+        self.layout().addWidget(self.table_wdg.native)
+
+    def _on_system_cfg_loaded(self):
+        self._populate_table()
```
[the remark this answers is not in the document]
fdrgsp: for now it is only called when the cfg is loaded

micromanager_gui/_gui_objects/_group_preset_table_widget.py
fdrgsp: We can get rid of it at the moment. I was thinking to use it for something later on but probably it won't be necessary.

micromanager_gui/_gui_objects/_group_preset_table_widget.py
```diff
@@ -0,0 +1,120 @@
...
+        vh.setDefaultSectionSize(24)
+        self.native.setEditTriggers(QtW.QTableWidget.NoEditTriggers)
+
+
+class MMGroupPresetTableWidget(QtW.QWidget):
+    def __init__(self):
+        super().__init__()
+
+        self._mmc = _core.get_core_singleton()
+        self._mmc.events.systemConfigurationLoaded.connect(self._populate_table)
```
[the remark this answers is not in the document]
ianhi: Definitely agree that we should be updating based on those signals. One thing to consider is what to do when we get a configGroupChanged event with a blank string for a preset. That can happen if the system state is changed such that it doesn't correspond to any preset (e.g. contains exposure and exposure was changed).  micromanager takes the approach of just putting a blank value in the dropdown which I always found to be confusing. In fact I actually thought it was a bug for a long time.

What do oyu think about doing the following when receiving a `configGroupChanged` signal with a blank string for the preset:

Change the visual indicators on the preset combobox to indicate that it **used** to be something but that that no longer exactly represents the state of the scope. I'm thinking something along the lines of graying out the text a little, or a red background, or maybe a strikethrough and also adding a tooltip that explains the situation. 

I wonder if we can be a bit more expressive in the UI here than micromanager is. In micromanager you can set a config group, but then if you change a property so that the system state no longer reflects that preset then the preset dropdown goes blank which was always a source of confusion for me. An easy way to end up there is if you put the exposure value into a preset. I now know that you shouldn't, but I suspect I'm not alone in having done that. The consequence is that if then change the exposure the configGroupChagned` signal will fire with an empty string as the preset.

So there's this splitting of state between what the user set, and what the actual state of the system is. I think we can represent that by instead of just showing preset names.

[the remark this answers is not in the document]
ianhi: See also the discussion of the two config signals here: https://github.com/tlambert03/pymmcore-plus/pull/38#issue-962948450 I think this is the relevant use case of the configSet signal. You set the config, but due to trickiness of what the devices do inside themselves the actual changed signal that gets fired is blank. Unless we are also listening to the set signal then you can up with the bad user experience of selecting something from a dropdown and having it immediately switch to blank.

micromanager_gui/_gui_objects/_group_preset_table_widget.py
```diff
@@ -0,0 +1,98 @@
...
+
+        # use only the first preset since device
+        # and property are the same for the presets
+        device, property, _, dev_prop_val_count = self._get_cfg_data(group, presets[0])
+
+        if len(presets) > 1 or dev_prop_val_count > 1:
+            # return PresetsWidget(group, text_color="white")
+            return PresetsWidget(group)
+        else:
+            return PropertyWidget(device, property)
```
ianhi, line 98: I have the same concerns here as with the PresetWidget PR, only checking the first one isn't safe. Is there any reason not to just have this become:

```suggestion
        return PresetsWidget(group)
```

fdrgsp: @ianhi 
> I have the same concerns here as with the PresetWidget PR, only checking the first one isn't safe. Is there any reason not to just have this become: ...

When I made this I was thinking that each presets should have the exact same properties...However I think that here we need to distinguish between `PresetsWidget()` and `PropertyWidget()`. `PresetsWidget()` is when we have _more then one property_ in the preset. Instead, If we _only have one_ (e.g. if I just want to have a light control slider or just the combobox for the binning) we want to create a `PropertyWidget()`.

ianhi: I'm a little bit confused, In particular by your slider group:

```
# Group: _slider_test
# Preset: NewPreset
ConfigGroup,_slider_test,NewPreset,Camera,TestProperty1,0.0000
```

I don't see why you would make that into a config group. As I see it the purpose of a config group is to give convenient named settings of potentially continuous things (and of course to group them together). So a more natural group would be:
```
# Group: _slider_test
# Preset: NewPreset
ConfigGroup, continuous_property,Low,Camera,TestProperty1,0.0

ConfigGroup, continuous_property,Med,Camera,TestProperty1,0.3

ConfigGroup, continuous_property,High,Camera,TestProperty1,0..8
```


> If we only have one (e.g. if I just want to have a light control slider or just the combobox for the binning) we want to create a PropertyWidget().

I don't think that a slider really represents the concept of presets well, in this case you should instead use the property browser rathre than shoehorning this into the groups and presets

ianhi: It seems like what would maybe be a better solution is a new tab where you can select arbitrary properties to include. So that you have the power of the proprety browser but customized to only the properties that are relevant.

fdrgsp: > I don't see why you would make that into a config group

something like this  `ConfigGroup,_slider_test,NewPreset,Camera,TestProperty1,0.0000` in the cfg of micromanager is considers it as a `ConfigGroup`. So I think it makes sense to create a `QSlider` (PropertyWidget()) instead of a QCombobBox (`PresetsWidget()`).

> So a more natural group would be:
  ```
  # Group: _slider_test
  # Preset: NewPreset
  ConfigGroup, continuous_property,Low,Camera,TestProperty1,0.0

  ConfigGroup, continuous_property,Med,Camera,TestProperty1,0.3

  ConfigGroup, continuous_property,High,Camera,TestProperty1,0..8
  ```

I agree with you on this and since in this group there is more than one `preset`, each `preset` is considered/created as a `PresetsWidget()`.

@ianhi Don't you agree? maybe I'm missing something...

fdrgsp:
> here's a very minor annoyance with the dropdown. I'm not even really sure how to fix this, but I suspect it will bug you too and maybe you'll figure it out 😂

@tlambert03 I think I fixed this annoying issue with [_set_combo_view](https://github.com/tlambert03/napari-micromanager/pull/122/commits/62ef957e7ce9eae6be36578a73297a316c4dba3d) 😂

tlambert03:
I see you moved the tab order back :)  ... I'm gonna move it again.  I don't like having the snap tab and the MDA tab (both related to acquisition, just different n dims), separated by a config-like tab.  It's the least like the other tabs and I think it needs to be over on the right.  maybe later we place it elsewhere entirely?

tlambert03 (approved):

PR #123, opened 2022-03-03
Generic PropertyWidget

tlambert03:
General purpose `PropertyWidget`.  takes a Device name and a property label and creates an appropriate widget (either a slider for int/float, a checkbox for booleans, combobox for options, or lineEdit for strings).

Can be used to compose more complicated device widgets.  It's essentially a replacement for the magicgui based `get_editor_widget` code in prop_browser (which can be rewritten as well).  It doesn't use magicgui, and is much better tested for two-way communication as a view onto an mmcore Property


micromanager_gui/_gui_objects/_property_widget.py
```diff
@@ -0,0 +1,71 @@
...
+            self._value_widget = QLabeledSlider(orientation)
+        elif allowed := self._mmc.getAllowedPropertyValues(device_label, prop_name):
+            self._value_widget = QComboBox()
+            self._value_widget.addItems(allowed)
+            self._value_widget.setValue = self._value_widget.setCurrentText
+            self._value_widget.value = self._value_widget.currentText
+        else:
+            self._value_widget = QLineEdit()
+            self._value_widget.setValue = self._value_widget.setText
+            self._value_widget.value = self._value_widget.text
```
fdrgsp, line 47: @tlambert03 should we add something like `self._value_widget.changed =` to each widget? (e.g. if it is a Slider `self._value_widget.changed = self._value_widget.valueChanged`).

This way we can add a new method that is called when the widget wall is changed:
```
def _on_wdg_changed(self, value: Any):
        self._value_widget.setValue(value)
```
and connect it like that:
`self._value_widget.changed.connect(self._on_wdg_changed)`

tlambert03: see the update I just pushed... it establishes a `PPropValueWidget` protocol that makes sure they all have the same interface

tlambert03:
This is mostly working, but I think there's probably a few edge cases that could be caught and handled better... such as what we do when core rejects the entered number... or what default range to use on a slider when no range is provided (we probably need to use a spinbox there)

tlambert03:
additional note, individual properties from state devices work a little better now.  These are properties for which `getAllowedPropertyValues` returns an empty tuple.  but which there _are_ indeed allowed values.  Fixed upstream in https://github.com/tlambert03/pymmcore-plus/pull/105

micromanager_gui/_core_widgets/_property_widget.py
```diff
@@ -0,0 +1,344 @@
...
+
+    def value(self) -> str:
+        return self.currentText()
+
+    def setValue(self, value: str) -> None:
+        value = str(value)
+        if value not in self._allowed:
+            raise ValueError(f"{value!r} must be one of {self._allowed}")
+        self.setCurrentText(value)
+
```
fdrgsp, line 138: @tlambert03 
can we add a method like this to the combobox?
```
def getValues(self) -> list:  # or def values(self)
        return[
            self.itemText(i) for i in range(self.count())
        ]
```

tlambert03: we could, but what we really care about is the model (the cmmcore instance) not the view (the widget).   So, if you want a convenience method to get the allowed values from core (using `core.getAllowedPropertyValues` or `core.getStateLabels`), I'd propose that we make a convenience for accessing allowable values from core on the `PropertyWidget` class itself, rather than the getting the current values in the combobox.  sound ok?

tlambert03: added `PropertyWidget.allowedValues`

tlambert03:
OK, I think this one is in pretty good shape and will be valuable in other places.  Merging after tests pass

PR #128, opened 2022-03-03
Generic PresetWidget

fdrgsp:
Similar to #123 but to generate `QComboBox` for micro-manager `groups` and `presets`.


tlambert03 (commented):
looks good!  tests please

fdrgsp:
@tlambert03 I think you can have a look...

micromanager_gui/_core_widgets/_presets_widget.py
```diff
@@ -0,0 +1,65 @@
...
+            raise ValueError(f"{self._group} group does not exist.")
+
+        self._presets = list(self._mmc.getAvailableConfigs(self._group))
+
+        if not self._presets:
+            raise ValueError(f"{self._group} group does not have presets.")
+
+        self._combo = QComboBox()
+        with signals_blocked(self._combo):
+            self._combo.addItems(self._presets)
```
tlambert03, line 34:
```suggestion
        self._combo.addItems(self._presets)
```

in this case, you don't need it because you haven't yet connected the combo box to core (that happens on line 39)

tests/core_widgets/test_presets_widget.py
```diff
@@ -0,0 +1,33 @@
...
+
+    groups = list(global_mmcore.getAvailableConfigGroups())
+
+    for group in groups:
+
+        presets = list(global_mmcore.getAvailableConfigs(group))
+
+        if len(presets) <= 1:
+            return
+
```
tlambert03, line 16:
```suggestion
```
any reason not to test this?  If `group` is in fact a valid configuration group name, then the user shouldn't have to always check whether that group has greater than preset before creating `PresetsWidget(group)`.  It should just show an empty (or 1-item) combobox.

tlambert03: Along those lines, we could consider handling the systemConfigurationLoaded event here and refresh the combo box choices (or, if the group is no longer present, disable the combobox and add a stub item that says "No Configuration group named...")

The goal for all of these "generic" widgets is to be as accurate of a view onto the state of mmcore as possible.  We can decide how we'd like to handle incompatible changes to core (such as the configuration group disappearing)... but one way or the other these widgets should either reflect the state of core at all times, or fail loudly.

Occasionally, we'll run into situations where mmcore itself simply doesn't emit the signal(s) we need to stay in sync, in which case we can try to fix upstream.  But short of that we should evaluate all the possible model changes that might necessitate a view change.  (This can also be an iterative process, and doesn't need to happen all in one PR)

fdrgsp: > any reason not to test this?

I put this here just for the tests...since I wanted to test both the mmcore.events.configSet signal 
```
with qtbot.waitSignal(global_mmcore.events.configSet):
            global_mmcore.setConfig(group, presets[1])
        assert wdg.value() == presets[1]
```
and the signal emitted by the widget

```
wdg.setValue(presets[0])
assert global_mmcore.getCurrentConfig(group) == presets[0]
```
And I thought that if there is only 1 preset (e.g. in the `System` group configuration) I could not use `preset[1]`


Do you think this is better:

```
def test_preset_widget(qtbot, global_mmcore: CMMCorePlus):

    groups = list(global_mmcore.getAvailableConfigGroups())

    for group in groups:

        presets = list(global_mmcore.getAvailableConfigs(group))

        wdg = PresetsWidget(group)
        qtbot.addWidget(wdg)
        wdg.show()

        assert list(wdg.allowedValues()) == presets

        p = presets[0] if len(presets) == 1 else presets[1]

        with qtbot.waitSignal(global_mmcore.events.configSet):
            global_mmcore.setConfig(group, p)
        assert wdg.value() == p

        wdg.setValue(presets[0])
        assert global_mmcore.getCurrentConfig(group) == presets[0]

        wdg._disconnect()
        # once disconnected, core changes shouldn't call out to the widget
        if len(presets) > 1:
            global_mmcore.setConfig(group, p)
            assert global_mmcore.getCurrentConfig(group) != wdg.value()
```

tlambert03: > And I thought that if there is only 1 preset (e.g. in the System group configuration) I could not use preset[1]

true, but you can handle that farther down.  In other words, first create a widget (and make sure the widget itself can be created.  If you determine that the widget won't have any more than 1 option, then yes, it doesn't make sense to test changing the option.

tlambert03: >         p = presets[0] if len(presets) == 1 else presets[1]

this still will give an index error in your test if there are no presets.  Do something like this:

```python
        wdg = PresetsWidget(group)
        qtbot.addWidget(wdg)
        presets = global_mmcore.getAvailableConfigs(group)
        assert list(wdg.allowedValues()) == presets
        if len(presets) < 1:
            return # no need testing the changes of a config group that has <1 item

        # change to the last preset
        with qtbot.waitSignal(global_mmcore.events.configSet):
            global_mmcore.setConfig(group, presets[-1])
        assert wdg.value() == presets[-1] == global_mmcore.getCurrentConfig()

```

fdrgsp: @tlambert03 

> Along those lines, we could consider handling the systemConfigurationLoaded event here and refresh the combo box choices (or, if the group is no longer present, disable the combobox and add a stub item that says "No Configuration group named...")

I definitely agree! I tested something like this and it work well when a new cfg is loaded.

```
def refresh(self) -> None:
        with signals_blocked(self._combo):
            self._combo.clear()
            if self._group not in self._mmc.getAvailableConfigGroups():
                self._combo.addItem(f'No group named {self._group}.')
                self._combo.setEnabled(False)
            else:
                presets = self._mmc.getAvailableConfigs(self._group)
                self._combo.addItems(presets)
                self._combo.setEnabled(True)
```

However I think it would be good to also have signals emitted when a `group` is `deleted` or when a `preset` is `deleted`. What do you think? Do you think we can add them in `pymmcore-plus`?

fdrgsp:
@tlambert03 I have also added this `_highlight_if_prop_changed` method. I think it can be a starting idea but I wanted to know what you think.

It is linked to the `PropertyChanged` signal and it highlights the combobox text in magenta if a property contained in the currently selected preset is changed (similar to micromanager when the combobox becomes empty)...

https://user-images.githubusercontent.com/70725613/156903048-db01086e-d40c-4e72-bdc0-a1b0ae06bf96.mov

ianhi:
> It is linked to the PropertyChanged signal and it highlights the combobox text in magenta if a property contained in the currently selected preset is changed (similar to micromanager when the combobox becomes empty)...

amazing - I like it a lot more than it going blank. Could also consider adding  we also add a tooltip explaining whats happening  so people won't be confused

tlambert03:
> @tlambert03 I have also added this _highlight_if_prop_changed method. I think it can be a starting idea but I wanted to know what you think.

I do like it!  but since it's not specific to a preset widget, it's a bit of PR feature-creep and I think it should go in another PR.  I can imagine us wanting to re-use this pattern elsewhere (show that an error has occurred), so i don't think it should be a method on this specific widget.  Let's address in another PR

fdrgsp:
> I do like it! but since it's not specific to a preset widget, it's a bit of PR feature-creep and I think it should go in another PR. I can imagine us wanting to re-use this pattern elsewhere (show that an error has occurred), so i don't think it should be a method on this specific widget. Let's address in another PR

I moved the `_set_wdg_color` method in `_util.py`. It can be used with any QWidget and it changes the color.

ianhi:
> I do like it! but since it's not specific to a preset widget, it's a bit of PR feature-creep and I think it should go in another PR. I can imagine us wanting to re-use this pattern elsewhere (show that an error has occurred), so i don't think it should be a method on this specific widget. Let's address in another PR


:+1: this would fit really naturally into #144. Otherwise @fdrgsp is this ready for a full review?

fdrgsp:
Yes you can have a look!👍🏻

micromanager_gui/_core_widgets/_presets_widget.py
```diff
@@ -0,0 +1,128 @@
...
+            _set_wdg_color(self.text_color, self._combo)
+
+    def _highlight_if_prop_changed(self, device: str, property: str, value: str):
+        """Set the text color to magenta if a preset property has changed"""
+
+        try:
+            dev_prop = [
+                (k[0], k[1])
+                for k in self._mmc.getConfigData(self._group, self._presets[0])
+            ]
```
ianhi, line 107: Any reason this can't live in init? Should be the same every time.

micromanager_gui/_core_widgets/_presets_widget.py
```diff
@@ -0,0 +1,128 @@
...
+
+        self.setLayout(QHBoxLayout())
+        self.layout().setContentsMargins(0, 0, 0, 0)
+        self.layout().addWidget(self._combo)
+        _set_wdg_color(self.text_color, self._combo)
+        self._combo.currentTextChanged.connect(self._on_combo_changed)
+        self._combo.textActivated.connect(self._on_text_activate)
+
+        self._mmc.events.configSet.connect(self._on_cfg_set)
+        self._mmc.events.configSet.connect(self._highlight_if_preset_state_changed)
```
ianhi, line 47: This should be changed to be `configChanged` and check whether we are in a valid config state.

micromanager_gui/_core_widgets/_presets_widget.py
```diff
@@ -0,0 +1,128 @@
...
+                self._combo.setEnabled(False)
+            else:
+                presets = self._mmc.getAvailableConfigs(self._group)
+                self._combo.addItems(presets)
+                self._combo.setEnabled(True)
+
+    def _highlight_if_preset_state_changed(self, group: str, preset: str):
+        """Set the text color to magenta if a preset has changed"""
+        if group != self._group and not self._mmc.getCurrentConfig(self._group):
+            _set_wdg_color("magenta", self._combo)
```
ianhi, line 96: What's meant to be happening here? I'm having trouble following, it seems like we set the color if some other config group has it's preset set?

ianhi (changes requested):
Ok this is off to a strong start! But I think we can simplify and make the invalidation of state a bit more robust. Right now you're listening to the `configSet` and `propretyChanged` signals. Instead you can just use the `configGroupChanged` signal. In that method it will either give you a string as the preset, in which case you can just swap what's displayed, or an empty string. For an empty string you would then color the text to show that it was the last state we were in, but aren't anymore

micromanager_gui/_core_widgets/_presets_widget.py
```diff
@@ -0,0 +1,128 @@
...
+        self.layout().setContentsMargins(0, 0, 0, 0)
+        self.layout().addWidget(self._combo)
+        _set_wdg_color(self.text_color, self._combo)
+        self._combo.currentTextChanged.connect(self._on_combo_changed)
+        self._combo.textActivated.connect(self._on_text_activate)
+
+        self._mmc.events.configSet.connect(self._on_cfg_set)
+        self._mmc.events.configSet.connect(self._highlight_if_preset_state_changed)
+        self._mmc.events.systemConfigurationLoaded.connect(self.refresh)
+        self._mmc.events.propertyChanged.connect(self._highlight_if_prop_changed)
```
ianhi, line 49: sorry this is the one that can be redone wiht the configchanged signal

fdrgsp:
> Instead you can just use the `configGroupChanged` signal.

When I started making this PR I tried to use the `configGroupChanged` signal but it did not work well...probably I did something wrong...

ianhi:
> configGroupChanged signal but it did not work well.

do you remember the issue? I think that a combination of that wiht the `setSignal` could be good

micromanager_gui/_core_widgets/_presets_widget.py
fdrgsp: @ianhi  Yes, that's what I wanted to do:

group A -> `preset a `-> (dev1 prop1 `val1a`), (dev2 prop2 val2a), ...
group B -> `preset b `-> (dev1 prop1 `val1b`), ...

if `preset b `is set, and `val1b != val1a`, then `preset a` will be colored in magenta

ianhi: ah - I thnk that the core `configGroupChanged` already does that logic for you :)

fdrgsp:
> do you remember the issue? I think that a combination of that wiht the `setSignal` could be good

I remember that it was not changing/sending the signal for all the presets...but I need to work on it again to remember.

So you are suggesting to only use the `configGroupChanged` and the `configChanged` signals instead of the `configSet` and `propertyChanged` signals, correct?

fdrgsp:
@ianhi I tried to do the following and it might be good if you repeat this as well. I'm not sure if I'm doing something wrong but as I mentioned above I don't get any configGroupChanged when I change properties...

Add this in the __init__
  ```
 @self._mmc.events.configGroupChanged.connect
   def _test(group: str, preset: str):
       print('group:', group, "preset:", preset)
```

and then in ipython (or where you normally do that): 

```
from micromanager_gui._core import get_core_singleton
from micromanager_gui._core_widgets._presets_widget import PresetsWidget
from pathlib import Path

mmc = get_core_singleton()
mmc.loadSystemConfiguration()

p = PresetsWidget("Channel")
p.show()

mmc.setConfig("Channel", "DAPI")
```
the `Channel` preset `DAPI` has `'Dichroic', 'Label', '400DCLP'`

if you change this property:
`mmc.setProperty('Dichroic', 'Label', 'Q585LP') `

there is no configGroupChanged signal emitted (DAPI in the combobox turns magenta because of `_highlight_if_prop_changed` method).

**Comment by ianhi** [comment]
ahhhh @fdrgsp  you're totally right - my bad!

The behavior of that signal is much more complicated than I remembered. There's some decently long discussion of this over here: https://github.com/micro-manager/mmCoreAndDevices/issues/25 Re-reading that and looking at the [implementation](https://github.com/micro-manager/mmCoreAndDevices/blob/0e5ba3c139ad5fab40aeec58bdd57c91f3c9c474/MMCore/CoreCallback.cpp#L479-L509) I still think that there needs to be a new signal added in the C++. It seems that the device adapters are responsible for calling the function that triggers this signal and they don't do that consistently. e.g. as you noticed with the demo camera it only seems to trigger for 3 of it's properties. This could be a nice opportunity to prototype a signal in pymmc+ and eventually upstream it to core.

Having gone down that rabbit I agree that your approach of checking the properties is the best way to detect a preset change. There's maybe a little bit of question as to where it should live (maybe on core?) but for this PR at least I think it's fine to have each preset widget listen.

Though I do think that your loop needs two things still:
1. Check all presets for the list of devices
2. When the property changes we should not always color the text. Sometimes a property change will swap us to a valid config group. In that case we should change the displayed version to reflect that.

For the logic of 2 it may be helpful to look at the mmcore [implementation](https://github.com/micro-manager/mmCoreAndDevices/blob/0e5ba3c139ad5fab40aeec58bdd57c91f3c9c474/MMCore/CoreCallback.cpp#L479-L509)

You can capture this behavior in a test (with changing the comments to appropriate asserts) by building on your existing test for that group:

```python
core.setConfig("Camera", "HighRes") # we match HighRes preset
# text: HighRes
# color: black
core.setProperty("Camera", "Binning", "2 # now we don't match any preset
# text: HighRes
# color: magenta
core.setProperty("Camera", "BinDepth", "10") # now we will fully match the MedRes preset
# text: MedRes
# color: black
```

tests/core_widgets/test_presets_widget.py
ianhi: > However I think it would be good to also have signals emitted when a group is deleted or when a preset is deleted. What do you think? Do you think we can add them in pymmcore-plus?

I think this would be a really set of signals to open and issue on `mmCoreAndDevices` about. You want to have a go at that? Feel free to tag me at least so I can follow along

fdrgsp:
> Though I do think that your loop needs two things still:
> - Check all presets for the list of devices
> - When the property changes we should not always color the text. Sometimes a property change will swap us to a valid config group. In that case we should change the displayed version to reflect that.

```
core.setConfig("Camera", "HighRes") # we match HighRes preset
# text: HighRes
# color: black
core.setProperty("Camera", "Binning", "2 # now we don't match any preset
# text: HighRes
# color: magenta
core.setProperty("Camera", "BinDepth", "10") # now we will fully match the MedRes preset
# text: MedRes
# color: black
```

@ianhi  I think I managed to get this nice behavior! Not sure if it is the best way to achieve that though!

micromanager_gui/_gui_objects/_camera_widget.py
```diff
@@ -15,7 +15,8 @@ def __init__(self):
         self.cam_roi_combo = QtW.QComboBox()
         self.crop_btn = QtW.QPushButton("Crop")
         self.px_size_spinbox = QtW.QDoubleSpinBox()
-        self.px_size_spinbox.setMinimum(1.0)
+        self.px_size_spinbox.setMinimum(0.0)
+        self.px_size_spinbox.setValue(0.0)
```
ianhi, line 19: Does this need to be part of this PR? it seems unrelated

micromanager_gui/_core_widgets/_presets_widget.py
```diff
@@ -0,0 +1,139 @@
...
+
+        if self._group not in self._mmc.getAvailableConfigGroups():
+            raise ValueError(f"{self._group} group does not exist.")
+
+        self._presets = list(self._mmc.getAvailableConfigs(self._group))
+
+        if not self._presets:
+            raise ValueError(f"{self._group} group does not have presets.")
+
+        self.dev_prop = get_dev_prop(self._group, self._presets[0])
```
ianhi, line 36: Ideally should check all presets. e.g. consider this (pathological) config group:

```
# Configuration presets
# Group: Camera
# Preset: MedRes
ConfigGroup,Camera,MedRes,Camera,Binning,2

# Preset: HighRes
ConfigGroup,Camera,HighRes,Camera,Binning,1
ConfigGroup,Camera,HighRes,Camera,BitDepth,12

# Preset: LowRes
ConfigGroup,Camera,LowRes,Camera,Binning,4
```

where bitdepth is in only one of the presets.

micromanager_gui/_core_widgets/_presets_widget.py
```diff
@@ -0,0 +1,139 @@
...
+        self._mmc.events.systemConfigurationLoaded.connect(self.refresh)
+        self._mmc.events.propertyChanged.connect(self._highlight_if_prop_changed)
+
+        self.destroyed.connect(self.disconnect)
+
+    def _on_text_activate(self, text: str):
+        """Used in case there is only one preset"""
+        if len(self._presets) == 1:
+            self._mmc.setConfig(self._group, text)
+            set_wdg_color(self.text_color, self._combo)
```
ianhi, line 59: This is relatively important, should get covered in a test

micromanager_gui/_core.py
```diff
@@ -59,11 +59,6 @@ def update_pixel_size(
     mmc : Optional[CMMCorePlus]
         optional mmcore object, by default `CMMCorePlus.instance()`
     """
-    if pixel_size == 1.0:
-        # TODO: this was previous behavior, but does this make sense?
-        # presumably it was here because the default value of the spinbox is 1.0?
-        return
-
```
ianhi, line 66: same commant as in camera_widget. Is this change necessary for the PR?

micromanager_gui/_util.py
```diff
@@ -170,3 +173,19 @@ def __init__(
...
+def get_dev_prop_val(
+    group: str, preset: str, mmcore: Optional[CMMCorePlus] = None
+) -> list:
+    mmc = mmcore or get_core_singleton()
+    return [(k[0], k[1], k[2]) for k in mmc.getConfigData(group, preset)]
+
+
+def get_dev_prop(group: str, preset: str, mmcore: Optional[CMMCorePlus] = None) -> list:
+    mmc = mmcore or get_core_singleton()
+    return [(k[0], k[1]) for k in mmc.getConfigData(group, preset)]
```
ianhi, line 191: these need docstrings explaining what they do, looking at them quickly I'm not totally sure I understand.. You can feel free to assume that the reader is knowledgeable about the internals of napari-micro when writing them.

micromanager_gui/_core_widgets/_presets_widget.py
```diff
@@ -0,0 +1,139 @@
...
+            for i in dpv:
+                cache_prop = self._mmc.getPropertyFromCache(i[0], i[1])
+                if cache_prop != i[2]:
+                    _set_combo = False
+                    break
+            if _set_combo:
+                with signals_blocked(self._combo):
+                    self._combo.setCurrentText(p)
+                    set_wdg_color(self.text_color, self._combo)
+                    return
```
ianhi, line 131: `get_dev_prop_val` is only used here and is just one line so we can inline it and also use some more explicit variable names (to help reviewers like me ;) )

```suggestion
        for preset in self._presets:
            _set_combo = True
            for (dev, prop, config_value) in self._mmc.getConfigData(self._group, p):
                cache_value = self._mmc.getPropertyFromCache(dev, prop)
                if cache_value != config_value:
                    _set_combo = False
                    break
            if _set_combo:
                with signals_blocked(self._combo):
                    self._combo.setCurrentText(preset)
                    set_wdg_color(self.text_color, self._combo)
                    return
        # None of the presets match the current system state
        set_wdg_color("magenta", self._combo)
```


You can tighten up this logic a bit. Also using the redoing of the loop in the comment above it's nice to use more explicit names that `i[0], i[1]`

micromanager_gui/_core_widgets/_presets_widget.py
```diff
@@ -0,0 +1,139 @@
...
+
+        if (device, property) not in self.dev_prop:
+            if self._mmc.getDeviceType(device) != DeviceType.StateDevice:
+                return
+            # a StateDevice has also a "Label" property. If "Label" is not
+            # in dev_prop, we check if the property "State" is in dev_prop.
+            if (device, "State") not in self.dev_prop:
+                return
+
+        self._set_if_props_match_preset()
```
ianhi, line 116: I think you can combine this an `_set_if_props...` into one function, because all this does is a little prep work and then calls the other function.

micromanager_gui/_core_widgets/_presets_widget.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def _highlight_if_preset_state_changed(self, group: str, preset: str):
+        """Set the text color to magenta if a preset has changed"""
+
+        if group == self._group:
+            return
+
+        dp = get_dev_prop(group, preset)
+        if any(x for x in dp if x in self.dev_prop):
+            self._set_if_props_match_preset()
```
ianhi, line 103: Why do we need this function? It seems like the `_on_cfg_set` should be enough on it's own?

ianhi (commented):
this is getting there! I left some comments on things that confused me, and a few places where the logic can be cleaned up, but I think you've nailed the core functionality.

micromanager_gui/_core_widgets/_presets_widget.py
fdrgsp: @ianhi Do you mean that we should make sure that the each preset (from the same group) have the same (dev, prop)?

Like if the group `Camera` has 3 presets (`HighRes`, `MedRes`, `LowRes`), they all need to have the same properties (with different values):

```
# Group: Camera
# Preset: MedRes
ConfigGroup,Camera,MedRes,Camera,Binning,2

# Preset: HighRes
ConfigGroup,Camera,HighRes,Camera,Binning,1

# Preset: LowRes
ConfigGroup,Camera,LowRes,Camera,Binning,4
```

or 

```
# Group: Camera
# Preset: MedRes
ConfigGroup,Camera,MedRes,Camera,Binning,2
ConfigGroup,Camera,HighRes,Camera,BitDepth,10

# Preset: HighRes
ConfigGroup,Camera,HighRes,Camera,Binning,1
ConfigGroup,Camera,HighRes,Camera,BitDepth,12

# Preset: LowRes
ConfigGroup,Camera,LowRes,Camera,Binning,4
ConfigGroup,Camera,HighRes,Camera,BitDepth,8
```

ianhi: No exactly. I do think that people should do that, but I'm saying that to account for the times that that they don't we should check all the presets to get the list of all devices.

fdrgsp: I've added the `_check_if_presets_have_same_props` method to check if the presets have the same properties.

micromanager_gui/_core_widgets/_presets_widget.py
fdrgsp: @ianhi I was thinking that we need to keep the part below only for the `propertyChanged` signal. The `configSet` signal emits `group` and `preset` and not (`device`, `property`, `value`). I wanted to run the `_set_if_props_match_preset` method only if the (device, property) is included in the preset. What do you think?
```
`if (device, property) not in self.dev_prop:
            if self._mmc.getDeviceType(device) != DeviceType.StateDevice:
                return
            # a StateDevice has also a "Label" property. If "Label" is not
            # in dev_prop, we check if the property "State" is in dev_prop.
            if (device, "State") not in self.dev_prop:
                return
```

tests/core_widgets/test_presets_widget.py
```diff
@@ -0,0 +1,41 @@
...
+def test_preset_widget(qtbot, global_mmcore):
+    for group in global_mmcore.getAvailableConfigGroups():
+        wdg = PresetsWidget(group, text_color="white")
+        qtbot.addWidget(wdg)
+        presets = list(global_mmcore.getAvailableConfigs(group))
+        assert list(wdg.allowedValues()) == presets
+
+        # no need testing the changes of a config group that has <= 1 item
+        if len(presets) <= 1:
+            return
```
ianhi, line 13: We may not need a test of them changing. But the code path to create them needs to be executed. They could get a standalone test, or this could be an if <1 and else branch. But needs to be some test of them

micromanager_gui/_util.py
```diff
@@ -170,3 +173,21 @@ def __init__(
...
+def set_wdg_color(color: str, wdg: QWidget):
+    wdg.setStyleSheet(f"color: {color};")
+
+
+def get_dev_prop_val(
+    group: str, preset: str, mmcore: Optional[CMMCorePlus] = None
+) -> list:
+    """return a list with (device, property, values) for the selected group preset"""
+    mmc = mmcore or get_core_singleton()
+    return [(k[0], k[1], k[2]) for k in mmc.getConfigData(group, preset)]
```
ianhi, line 187: No longer used. Can remove

```suggestion
```

ianhi:
@fdrgsp  if you merge or rebase on main then the windows tests should start passing

micromanager_gui/_core_widgets/_presets_widget.py
```diff
@@ -0,0 +1,144 @@
...
+
+        self.destroyed.connect(self.disconnect)
+
+    def _check_if_presets_have_same_props(self):
+        for idx, preset in enumerate(self._presets):
+            if idx == 0:
+                continue
+            device_property = get_dev_prop(self._group, preset)
+            if len(device_property) != len(self.dev_prop):
+                raise ValueError(f"{self._presets} must have the same properties.")
```
ianhi, line 64: Can we turn this into a warning instead. Explaining that it will may not update as expected. Alternatively I'm not even sure we need to error/warn here. All we need to do is expand `dev_prop` to make it encompass all possible devices. You've already made `_set_if_props_match_preset` robust enough that it should handle that case :)

micromanager_gui/_util.py
```diff
@@ -170,3 +173,13 @@ def __init__(
...
+
+
+def set_wdg_color(color: str, wdg: QWidget):
+    wdg.setStyleSheet(f"color: {color};")
+
+
+def get_dev_prop(group: str, preset: str, mmcore: Optional[CMMCorePlus] = None) -> list:
+    """return a list with (device, property) for the selected group preset"""
+    mmc = mmcore or get_core_singleton()
+    return [(k[0], k[1]) for k in mmc.getConfigData(group, preset)]
```
ianhi, line 185: if you do it like this then I think we can eliminate `_check_if_presets_have_same_props`

```suggestion
def get_dev_prop(
    group: str, preset: str, mmcore: Optional[CMMCorePlus] = None
) -> List[Tuple[str, str]]:
    """
    Return a list of all (device, property) tuples used in the config group's presets
    """
    mmc = mmcore or get_core_singleton()
    dev_props = []
    for preset in mmc.getAvailableConfigs(group):
        dev_props.extend([(k[0], k[1]) for k in mmc.getConfigData(group, preset)])
    return dev_props
```

micromanager_gui/_core_widgets/_presets_widget.py
ianhi: My prefernce is:
1. not needing this (see below comment for how)
2. warning
3. error

fdrgsp: @ianhi but if we don't use `_set_if_props_match_preset` to warn the user that the presets are not correct, then what do we need it for?
The change in the `get_dev_prop` method you suggested below is only returning a list, where are we using it? I'm not sure I understand what you mean

ianhi: I'm sorry I'm not being more clear. It's totally possible that (like many times before) I'm actually missing something and asking you to implement something that isn't possible.... I'll try to mess around tonight and generate the test case I'm thinking of (and in process decide if what I'm thinking is on the right track)

ianhi (approved):
I have one concern but I'm happy to try to address it in a future PR rather than holding this nice one up any longer. So up to @tlambert03 to review - but gets a :+1:  from me.

My maybe follow up:

I figured out how to demonstrate the issue of device properties I was talking about. If you add this config group:

```
# Preset: HighRes
ConfigGroup,Camera-Kinda-Broken,HighRes,Camera,BitDepth,12

# Preset: MedRes
ConfigGroup,Camera-Kinda-Broken,MedRes,Camera,Binning,2
ConfigGroup,Camera-Kinda-Broken,MedRes,Camera,BitDepth,10

# Preset: LowRes
ConfigGroup,Camera-Kinda-Broken,LowRes,Camera,Binning,4
ConfigGroup,Camera-Kinda-Broken,LowRes,Camera,BitDepth,8
```

and then run this script:
```python
from micromanager_gui._core import get_core_singleton
from micromanager_gui._core_widgets._presets_widget import PresetsWidget
from pathlib import Path
from IPython import embed

mmc = get_core_singleton()
mmc.loadSystemConfiguration(Path(__file__).parent/"tests"/"test_config.cfg")
from qtpy.QtWidgets import QApplication, QMainWindow, QVBoxLayout, QWidget, QPushButton
from qtpy import QtCore

app = QApplication([])

main = QWidget()
layout = QVBoxLayout()
p = PresetsWidget("Camera-Kinda-Broken")
b_med = QPushButton("med res binning")
b_low = QPushButton("low res binning")

layout.addWidget(p)
layout.addWidget(b_med)
layout.addWidget(b_low)
mmc.setConfig("Camera-Kinda-Broken", "MedRes")

def bin_med():
    mmc.setProperty("Camera", "Binning", 2)
def bin_low():
    mmc.setProperty("Camera", "Binning", 4)
b_med.clicked.connect(bin_med)
b_low.clicked.connect(bin_low)


main.setLayout(layout)
# embed(colors='neutral')

main.show()
# mmc.setConfig("Channel", "DAPI")
app.exec_()
```

then I'd argue that changing the binning should invalidate the current selection. However, the current behavior just raises an error on creation. With my proposed change  to the `get_dev_prop` function then on changing the binning this behaves as i'd expect: changing the binning triggers changes in what the widget displays.

ianhi:
Light bump here

tlambert03 (commented):
This looks good to me.

The only thing I don't like is the handling of styles.  We shouldn't have "text_color" as a parameter in the `__init__`.  I assume you added the `set_wdg_color` in response to my comment to generalize it, but simply pulling out "setStyleSheet" into a new function isn't quite what I meant.  (I meant more that we should have a generalized mechanism for error styles, but let's not do it like this).

Can you just remove `set_wdg_color` and remove the `text_color`?  One thing we need to check is that this still works when the theme changes (i.e. setting text to "black" assumes a very specific theme)

fdrgsp:
@tlambert03 I removed `set_wdg_color` and `text_color` and modify the code a little...

we are still setting the `_combo` text color to `magenta` if the preset is not "set", however the `_combo` text color is restore to `default` without using any `text_color` variable in the `__init__` by simply using `setStyleSheet("")` (e.g. in `napari` will be white when in dark mode and black when in light mode. I tested it and it works).

Do you think that's ok?

fdrgsp:
@ianhi I've added your suggestion on `get_dev_prop` but I renamed this method as `get_group_dev_prop`. 
I've also added a different method called `get_preset_dev_prop` that I have used to in `_check_if_presets_have_same_props` because I think it is important for a user to know if the `presets` do not have the same `properties`. However, instead of an `error`, I used a `warning`.

ianhi (approved):
Awesome! This now completely give the behavior i'd expect

tlambert03 (approved):
LGTM, thanks @fdrgsp!

PR #131, opened 2022-03-05
new QProperty browser

tlambert03:
This is a new (and I think improved) property browser widget, no longer using magicgui, but instead using the new PropertyWidgets from #123.  Some small changes were made to PropertyWidget in the process and a new Micro-Manager main menu item is added here.

Note, one thing that's still broken (always was probably) is reloading a system configuration when the browser is open.  Could try to fix in this PR, but will likely save for a followup


micromanager_gui/main_window.py
```diff
@@ -127,6 +126,24 @@ def __init__(self, viewer: napari.viewer.Viewer, remote=False):
...
+
+    def _patch_viewer_menu(self):
+        self._menu = QtW.QMenu("&Micro-Manager")
+
+        action = self._menu.addAction("Device Property Browser...")
+        action.triggered.connect(self._show_prop_browser)
+
+        v = getattr(self.viewer, "__wrapped__", self.viewer)
+        bar = v.window._qt_window.menuBar()
+        bar.insertMenu(list(bar.actions())[-1], self._menu)
```
ianhi, line 139: I've wondered how to do this before. maybe a good feature request for napari is a method to add these without accessing private variables like that (or know about wrapped)

ianhi: Can we pick up the napari theming so that the menu doesn't look out of place?

menu-themeing

ianhi:
Not quite sure what's happening here but for at least some propreties I think we need to delay setting the value in core until the user hits enter or leaves the textbox. For example setting the asyncpropertyleader (a string property) and typing fast grids everythign to a halt:
slow


I typed out: "typing here is very slow"

micromanager_gui/main_window.py
tlambert03: > maybe a good feature request for napari is a method to add these without accessing private variables like that (or know about wrapped)

it's [coming in npe2](https://github.com/napari/npe2/blob/main/npe2/manifest/menus.py).  don't tell anyone about this workaround :joy:

tlambert03: sure, can fix the theme

tlambert03:
> For example setting the asyncpropertyleader (a string property) and typing fast grinds the event loop to a halt:

thank you, will dig

tlambert03:
> For example setting the asyncpropertyleader (a string property) and typing fast grinds the event loop to a halt:

Actually... I can't find anything with which to test this!  is that a property of a real camera?  I changed the lineEdit signal to `editingFinished` ... meaning it won't update core until you hit the enter key, or focus away from the field.  Hopefully that helps

I could do a similar thing for sliders if we need to later

ianhi:
> Actually... I can't find anything with which to test this! is that a property of a real camera? 

Do you mean you don't have that property? I added it to the demo camera pretty recently so you may need to update micromanager

tlambert03:
> I added it to the demo camera pretty recently so you may need to update micromanager

aha! that's right, thanks :)

ianhi:
Setting that property now throws these errors in the console:

```
An error occurred while trying to format an error and show it in console.
You can try to uninstall IPython to disable rich traceback formatting
And/or report a bug to napari
```

also: the ideal behavior here is that I change the leader property and then the follower property should up after a delay (also a property) to be the same value.

micromanager_gui/_core_widgets/_property_widget.py
```diff
@@ -150,7 +174,7 @@ class StringWidget(QLineEdit):
 
     def __init__(self, parent: Optional[QWidget] = None) -> None:
         super().__init__(parent)
-        self.textChanged.connect(self.valueChanged.emit)
+        self.editingFinished.connect(self.valueChanged.emit)
```
ianhi, line 177:
```suggestion
        self.editingFinished.connect(lambda: self.valueChanged.emit(self.value()))
```

fixes the errors

micromanager_gui/_core.py
```diff
@@ -85,3 +85,11 @@ def update_pixel_size(
...
         mmc.setPixelSizeUm(resolutionID, pixel_size / mag)
         mmc.setPixelSizeConfig(resolutionID)
+
+
+def iter_dev_props(mmc: Optional[CMMCorePlus] = None) -> Iterator[Tuple[str, str]]:
+    """Yield all pairs of currently loaded (device_label, property_name)."""
+    mmc = mmc or get_core_singleton()
+    for dev in mmc.getLoadedDevices():
+        for prop in mmc.getDevicePropertyNames(dev):
+            yield dev, prop
```
ianhi, line 95: replace with the new pymmcore-plus method?

micromanager_gui/_core_widgets/_property_browser.py
```diff
@@ -0,0 +1,189 @@
...
+
+from .._core import get_core_singleton
+from ._property_widget import PropertyWidget
+
+
+def iter_dev_props(mmc: CMMCorePlus) -> Iterator[Tuple[str, str]]:
+    for dev in mmc.getLoadedDevices():
+        for prop in mmc.getDevicePropertyNames(dev):
+            yield dev, prop
+
```
ianhi, line 29: same comment about replace with pymmcore-plus method

tlambert03: lets wait for that to be released and then do a cleanup here

micromanager_gui/_core_widgets/_property_browser.py
```diff
@@ -0,0 +1,189 @@
...
+            item = QTableWidgetItem(f"{dev}-{prop}")
+            wdg = PropertyWidget(dev, prop, core=self._mmc)
+            self.setItem(i, 0, item)
+            self.setCellWidget(i, 1, wdg)
+            if wdg.isReadOnly():
+                # TODO: make this more theme aware
+                item.setBackground(QColor("#AAA"))
+                wdg.setStyleSheet("QLabel { background-color : #AAA }")
+
+        # TODO: install eventFilter to prevent mouse wheel from scrolling sliders
```
ianhi, line 70: This will definitely be nice. It also affects the spinBoxes it seems

ianhi (approved):
This is really nice to use :+1:

micromanager_gui/_core_widgets/_property_browser.py
tlambert03: yeah.  I tried for a while, but it kept failing.  So gave up for later! :)

Issue #130, opened 2022-03-05
Update for new pymmcore-plus mda interface 


tlambert03:
@ianhi, we need to update this repo now for https://github.com/pymmcore-plus/pymmcore-plus/issues/101

ianhi:
Indeed. I actually got started on this but then it began morphing into a full overhaul of the mda widget. Maybe worth it to have a smaller part just updating the signals?

tlambert03:
Yeah, a bare minimum compatibility pr would be good.  How does the widget need to change?

ianhi:
I was hoping to make it more standalone like all the other PRs. It's also currently really coupled to the explore-sample widget (which actually handles the display) and some of it's logic lives on the main window rather than the widget

tlambert03:
ah, yes.  That would be excellent

tlambert03:
i'm not sure if there's anything actionable here anymore?  @ianhi?

ianhi:
I believe all is well here

PR #132, opened 2022-03-05
update for new pymmcore-plus mda

ianhi:
Is it ok to just relying on the old tests for now? I'd like to add more once the mda widgets get a bit cleaned up as discussed in #130

### DIFF
--- a/micromanager_gui/multid_widget.py
+++ b/micromanager_gui/multid_widget.py
@@ -14,6 +14,7 @@ from useq import MDASequence
 
 if TYPE_CHECKING:
     from pymmcore_plus import RemoteMMCore
+    from pymmcore_plus.mda import PMDAEngine
 
 ICONS = Path(__file__).parent / "icons"
 
@@ -98,8 +99,8 @@ class MultiDWidget(QtW.QWidget, _MultiDUI):
         super().__init__(parent)
         self.setup_ui()
 
-        self.pause_Button.released.connect(self._mmc.toggle_pause)
-        self.cancel_Button.released.connect(self._mmc.cancel)
+        self.pause_Button.released.connect(lambda: self._mmc.mda.toggle_pause())
+        self.cancel_Button.released.connect(lambda: self._mmc.mda.cancel())
 
         # connect buttons
         self.add_pos_Button.clicked.connect(self.add_position)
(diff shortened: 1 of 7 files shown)


setup.cfg
```diff
@@ -30,7 +30,7 @@ include_package_data = True
 install_requires =
     napari >=0.4.13
     scikit-image
-    pymmcore-plus >=0.3.1
+    pymmcore-plus >=0.4.0
```
ianhi, line 33: this is presuming we bump the version numbers that much - which I think makes sense?

ianhi:
ahh. and all the tests will of course fail bc `0.4.0` doesn't exist

setup.cfg
tlambert03: for the purpose of this PR, let's set this to the main branch 
something like
```
install_requires =
  pymmcore-plus  @ git+https://github.com/tlambert03/pymmcore-plus
```

tlambert03: then once everything is settled, we can cut a new release and actually depend on 0.4.

ianhi:
Tests will still fail as remotemmcore does not support mda anymore

ianhi:
@tlambert03 thoughts on what to do about remote in the tests? 

Though also looks as though I've indeed broken some tests even for local

tlambert03:
> Tests will still fail as remote mmcore

yeah, and so now we basically need to give up on it over here too :)  I'm ok with getting rid of the remote parameter in the global_mmcore fixture.

ianhi:
ah this is not ready. I do need to add a basic test (it caught some errors)

ianhi:
Ok tests all passed locally and I added an engine swap to the existing test. Should be good for review once all tests pass here.

micromanager_gui/multid_widget.py
```diff
@@ -98,8 +99,8 @@ def __init__(self, mmcore: RemoteMMCore, parent=None):
         super().__init__(parent)
         self.setup_ui()
 
-        self.pause_Button.released.connect(self._mmc.toggle_pause)
-        self.cancel_Button.released.connect(self._mmc.cancel)
+        self.pause_Button.released.connect(lambda: self._mmc.mda.toggle_pause())
```
tlambert03, line 102: i'm curious, did you find that the lambda was necessary here for this to work?  I've definitely seen that before.  Always kinda wonder what the circumstances are though

tests/test_main_window.py
```diff
@@ -67,12 +68,20 @@ def test_saving_mda(qtbot: "QtBot", main_window: MainWindow, T, C, splitC, Z):
 
         mda = None
 
-        @main_window._mmc.events.sequenceStarted.connect
+        mmc = main_window._mmc
+        # re-register twice to fully exercise the logic of the update
+        # functions - the initial connections are made in init
+        # then after that they are fully handled by the _update_mda_engine
+        # callbacks
```
tlambert03, line 75: true, though perhaps we should have used `self.register_mda_engine(MDAEngine(self))` instead of `self._mda_engine = ... ` in the `__init__` method over in pymmcore_plus, to avoid any potential diversion between init and method-based registration in the future?

tlambert03 (approved):
looks good!  thanks.  Will merge now.  left a couple side thoughts, not important here

tlambert03:
note @fdrgsp, this is a breaking API change on the pymmcore-plus side.  After this merges, you will need to have the latest main branch of pymmcore-plus active in your environment.

PR #134, opened 2022-03-05
Begin cleanup and decoupling of mda

ianhi:
Building on #132 

Prior to this PR there was some complex coupling between the multi_d widget, the explore sample widget and the main window. The aim here is to simplify and break that up. My goals are:

1.  `explore_sample` and `mda_widget` do not have to know anything about napari  or eachother
      - They are each responsible only for constructing the appropriate MDASequence and  pause/stop
2. All interactions with Napari are confined to main_window widget
3. we also support displaying MDAs that initiated outside of napari-micromanager

(Partially done - but didn't want to lose this description that I wrote)

PR #139, opened 2022-03-07
Ensure a new Core instance for each test

ianhi:
Previously we were always getting the same core instance for each test due to the instancing of `CMMCorePlus.instance()`

this fixes that, and then fixes the test that then broke (which is also whats breaking #134 )

Pulling out of that to make things simpler to review

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,6 +4,7 @@ from typing import Tuple
 
 import numpy as np
 import pytest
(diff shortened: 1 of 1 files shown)


ianhi:
The test failure was because we were loading the default demo config - which has a misspelling in it: https://github.com/micro-manager/micro-manager/pull/1424

tlambert03 (approved):

Issue #136, opened 2022-03-06
pixelSizeConfiguration considerations


tlambert03:
Opening this issue to get some feedback from @fdrgsp on the design goals we should be having for maintenance of pixel size configuration in general.  We should also compare/contrast with how this would be done in micromanager (there's always merit in maintaining consistency between the two, unless we decide that we can definitely offer something better).

From what I can tell, this is the goal

1) we offer an input somewhere that lets the user enter their camera pixel size.
2) when either the camera pixel size _or_ the objective changes, we:
   **a.** try to glean what the magnification of the objective is from the objective label (this is definitely something not done in micromanager, and probably error prone... might be better to have an explicit API rather than do the magic).
   **b.** define a new pixelSizeConfig entry for the objective
   **c.** `setPixelSizeUm` using the mag and camera pixel
    **d.** set the config as current...

I definitely like the general goal here, but I think we need an independent object, probably one that doesn't really need widgets, which monitors core property changes for the objective and whatever means we use to retrieve camera pixel size.

I also don't think we need that pixel size entry to be part of the main window.

fdrgsp:
I definitely agree with that! I created the `px_size spinbox` at the beginning only to test the sample explorer widget. So I think we can remove that and for now rely only on setting the px size info in the system configuration file.

> retrieve camera pixel size.

how would you suggest to do that if there is no initial px size info provided by the system configuration file? (I'd love to get the camera px size info from the micromanager camera properties but there I don't think there is this info).

tlambert03:
> how would you suggest to do that if there is no initial px size info provided by the system configuration file? 

not sure.  I'm not opposed to having an input widget dialog somewhere.  But don't think it needs a permanent home in main window.  I'm adding a `Micro-Manager` menu in the #131, so perhaps we could add it as a command up there.

I'm going to merge #119 now, so explorer may break temporarily... but I definitely want to get it working again ASAP cause it's very cool.  Let's sync up and discuss how best to collect the necessary info

fdrgsp:
Done in `pymmcore-widgets` `PixelSizeWidget`.

PR #141, opened 2022-03-07
Add a widget to create pixel size configurations

fdrgsp:
This might be a temporary solution to set the pixel size. As @tlambert03 mentioned in #136 (I'm adding a `Micro-Manager` menu...), this widget could be placed in the menu and the user can define pixel size configs from there.

What do you think?

https://user-images.githubusercontent.com/70725613/157528636-7246d51a-b879-4e2e-83d0-857bacb75f56.mov


fdrgsp:
> What happens if you click `new row`? Am I right that in most cases there will only be one combobox of objectives? If that's the case then I think we should just display all the rows at once and you click around to choose which ones to edit.

@ianhi it's just my opinion but I think we should make the user decide how to set the configuration...in some cases they can add some optical components in the path that will be used only with a specific objective, etc...

tlambert03:
is it possible to directly modify image pixel size?  (doesn't look like it is yet).  I don't think the user should be "forced" to use the camera pixel/mag math...  If they know their image pixel size (which is the default interface in micromanager right?)  they should be allowed to enter it

fdrgsp:
@tlambert03 I think everything is working as it should now...

fdrgsp:
closing since it has been moved to https://github.com/pymmcore-plus/pymmcore-widgets.

tlambert03:
did this feature specifically get implemented?  If not, we shouldn't close the issue (but we can transfer it)

fdrgsp:
@tlambert03  yes sorry, I've added it to [pymmcore-widgets](https://github.com/pymmcore-plus/pymmcore-widgets).

tlambert03:
Can you link the pr here so we can see it?

fdrgsp:
@tlambert03 #141

tlambert03:
:joy: ... that links to _this_ PR.  looking through your history, I'm guessing it was https://github.com/pymmcore-plus/pymmcore-widgets/pull/6

If I recall, there were definitely some UI and organization-related questions remaining to be answered in this PR.  Let's not merge un-discussed, un-reviewed things into the new repo.  🙏

fdrgsp:
😂 sorry I didn't get that. yes that's the one. do you want me to reopen it in [pymmcore-widgets](https://github.com/pymmcore-plus/pymmcore-widgets)?

tlambert03:
No, just create issues reminding us as a group to evaluate them

PR #142, opened 2022-03-09
Stage widget

tlambert03:
creating a new General purpose StageDevice Widget


tlambert03:
still very much WIP, but managed to get rid of 1400 lines of code and have this so far:

Untitled

tlambert03:
this is getting closer to ready for a critical eye @fdrgsp.  Particularly we want to check whether most of the logic in main_window has been preserved (though it's all removed from main_window now).  might be worth checking this on a real scope.  Currently, it's rather generic, so it's also likely something specific to (e.g.) autofocus may have been lost.

Couple minor aesthetic issues remaining, but looking decent:

Untitled 2

micromanager_gui/_gui_objects/_xyz_stages.py
```diff
@@ -1,69 +1,41 @@
...
+        self.setLayout(QHBoxLayout())
+        self.layout().setContentsMargins(0, 0, 0, 0)
+        self.layout().setSpacing(6)
+        self._mmc = _core.get_core_singleton()
+        self._on_cfg_loaded()
+        self._mmc.events.systemConfigurationLoaded.connect(self._on_cfg_loaded)
+
+    def _on_cfg_loaded(self):
+        self._clear()
+        if dev := self._mmc.getXYStageDevice():
```
ianhi, line 19: I think we should provide an object for generic `StageWidgets` and then a subclass that follows the default stage widget. Like what I did in #150 (in spirit if not in implementation)

tlambert03: that's primarily what this PR is doing.  See [`micromanager_gui/_core_widgets/_stage_widget.py`](https://github.com/tlambert03/napari-micromanager/pull/142/files#diff-2c246c921e3e8559c18cb076e6e01fbf676d34a9751d7a68dc9322187e790c7d)

this file here is just for the few things that are special about napari micromanager (like the QCollapsible).  hence why it's in gui_objects :)

tlambert03: and this one arranges the generic stage widget into a multi widget.  That could also happen in core, but the primary widget here is `class StageWidget(QWidget)` in that other file

ianhi:
One test failure will be fixed with a rebase - other one looks real

ianhi:
Renders squashed for me when I run the dev script:
image

gets fixed if i toggle the qcollapsible on and off - but broken on opening

tlambert03:
> gets fixed if i toggle the qcollapsible on and off - but broken on opening

yep, this needs https://github.com/napari/superqt/pull/72 to be released

micromanager_gui/_core_widgets/_stage_widget.py
```diff
@@ -0,0 +1,225 @@
...
+    def _move_stage(self, x, y):
+        if self._dtype is DeviceType.XYStage:
+            self._mmc.setRelativeXYPosition(self._device, x, y)
+        elif self._dtype is DeviceType.AutoFocus:
+            self._mmc.setAutoFocusOffset(y)
+        else:
+            self._mmc.setRelativePosition(self._device, y)
+        self._mmc.waitForDevice(self._device)
+        if self._snap_on_click:
+            self._mmc.snap()
```
ianhi, line 218: Either there needs to be a change in core to have this stop live mode - or there there needs to be acheck here of if a sequence is running. This can result in this error:


```
untimeError                              Traceback (most recent call last)
File ~/Documents/oss/micro/napari-micromanager/micromanager_gui/_core_widgets/_stage_widget.py:207, in StageWidget._on_click(self=<micromanager_gui._core_widgets._stage_widget.StageWidget object>)
    205 btn: QPushButton = self.sender()
    206 xmag, ymag = self.BTNS[f"{PREFIX}.{btn.text()}"][-2:]
--> 207 self._move_stage(self._scale(xmag), self._scale(ymag))
        xmag = 2
        ymag = 0
        self = <micromanager_gui._core_widgets._stage_widget.StageWidget object at 0x7f5d33328ee0>

File ~/Documents/oss/micro/napari-micromanager/micromanager_gui/_core_widgets/_stage_widget.py:218, in StageWidget._move_stage(self=<micromanager_gui._core_widgets._stage_widget.StageWidget object>, x=20, y=0)
    216 self._mmc.waitForDevice(self._device)
    217 if self._snap_on_click:
--> 218     self._mmc.snap()
        self = <micromanager_gui._core_widgets._stage_widget.StageWidget object at 0x7f5d33328ee0>
        self._mmc = <CMMCorePlus at 0x7f5d3326cd10>

File ~/mambaforge/envs/pymm/lib/python3.10/site-packages/pymmcore_plus/core/_mmcore_plus.py:648, in CMMCorePlus.snap(self=<CMMCorePlus>, fix=True, *args=())
    629 def snap(self, *args, fix=True) -> np.ndarray:
    630     """
    631     snap and return an image.
    632 
   (...)
    646     img : np.ndarray
    647     """
--> 648     self.snapImage()
        self = <CMMCorePlus at 0x7f5d3326cd10>
    649     img = self.getImage()
    650     self.events.imageSnapped.emit(img)

File ~/mambaforge/envs/pymm/lib/python3.10/site-packages/wrapt/decorators.py:469, in synchronized.<locals>._synchronized(wrapped=<bound method CMMCorePlus.snapImage of <CMMCorePlus>>, instance=<CMMCorePlus>, args=(), kwargs={})
    463 @decorator
    464 def _synchronized(wrapped, instance, args, kwargs):
    465     # Execute the wrapped function while the original supplied
    466     # lock is held.
    468     with lock:
--> 469         return wrapped(*args, **kwargs)
        wrapped = <bound method CMMCorePlus.snapImage of <CMMCorePlus at 0x7f5d3326cd10>>
        args = ()
        kwargs = {}

File ~/mambaforge/envs/pymm/lib/python3.10/site-packages/pymmcore_plus/core/_mmcore_plus.py:553, in CMMCorePlus.snapImage(self=<CMMCorePlus>)
    551 @synchronized(lock)
    552 def snapImage(self) -> None:
--> 553     return super().snapImage()

RuntimeError: This operation can not be executed while sequence acquisition is running.
```

to reproduce that error:
1. launch
2. start live mode
3. trying moving stage

fdrgsp: I love this "snap_after_stage_moves" function but I think we can remove it for now and see how to add it back once the stage, the snap and the live button widgets are merged. Do you agree?

micromanager_gui/_core_widgets/_stage_widget.py
```diff
@@ -0,0 +1,225 @@
...
+    def __init__(self, levels=2, device=None, mmcore=None):
+        super().__init__()
+        self._mmc = mmcore or _core.get_core_singleton()
+        self._device = device or self._mmc.getXYStageDevice()
+        self._dtype = self._mmc.getDeviceType(self._device)
+        assert self._dtype in STAGE_DEVICES, f"{self._dtype} not in {STAGE_DEVICES}"
+        self.setStyleSheet(STYLE)
+        self._snap_on_click = True  # TODO: make widget
+
+        self._step = QSpinBox()
```
ianhi, line 82:
```suggestion
        self._step = QDoubleSpinBox()
```

currently is limited to 1 micron steps

ianhi:
I think there needs to be a UI element showing the difference between a single arrow move vs a double arrow move. As it stands I'm not actually sure of the difference.

ianhi:
> I think there needs to be a UI element showing the difference between a single arrow move vs a double arrow move. As it stands I'm not actually sure of the difference.


Ahh I see this in the tooltip so less urgent - but still it was not immediately obvious. I think MM behavior of allowing you to set two values is pretty nice. One for fine adjust, and one for something like moving a field of view over

fdrgsp:
Since the `AutoFocus` devices work in a slightly different way, I think It is best to have an `autofocus_stage_widget` (based on this `Stage widget`) specifically modified for the `Autofocus` devices.

fdrgsp:
@tlambert03 I think you can have a look. I will test it on s15 when the microscope will be free!

ianhi:
Is there anything more that needs to be done here?

fdrgsp:
> Is there anything more that needs to be done here?

I will soon open a new PR on a more complete version that takes into account also the `micromanager` `autofocus` behavior.

tlambert03 (commented):
I'm ok with this whenever you guys are.  We can follow up with autofocus stuff in another PR.  @fdrgsp, do you feel like this is complete as is?  or want more time with it?

(thanks again for taking this over the finish line!!)

ianhi:
> We can follow up with autofocus stuff in another PR

I'd vote for merging sooner as this will fix issues I have with a `z` stage that doesn't have `z` in the name so the value doesn't update on main :(

> I'm ok with this whenever you guys are.

Trying this out again since I last did I'm I do have one thing I dont love. There no longer seems to be a way to have the stage controls be part of the main controls layout - it's now triggered by a button:

image

and that leaves you with a floating stage controls:
image

I'm not sure exactly what is meant to be in "Light Sources" but I'm not sure it should be given equal screen space to the XYZ stage. Overall the layout in hte main widget and light source stuff  feels like a bit of feature creep. Could this be limited to just the new stage widget and then those other things get considered in followups?

micromanager_gui/_core_widgets/_stage_widget.py
```diff
@@ -0,0 +1,362 @@
...
+        self._mmc.events.systemConfigurationLoaded.connect(self._os_system_cfg)
+
+    def _enable_wdg(self, enabled):
+        self._step.setEnabled(enabled)
+        self._btns.setEnabled(enabled)
+        self.snap_checkbox.setEnabled(enabled)
+        self.radiobutton.setEnabled(enabled)
+        self._poll_cb.setEnabled(enabled)
+
+    def _os_system_cfg(self):
```
ianhi, line 204:
```suggestion
    def _on_system_cfg(self):
```
typo?

fdrgsp:
@tlambert03 
> I'm ok with this whenever you guys are. We can follow up with autofocus stuff in another PR. @fdrgsp, do you feel like this is complete as is? or want more time with it?

You can have look, I think this is complete!

However I'm not sure it will work well if there is an autofocus device in the cfg (e.g. Nikon PFS)...these devices in micromanager have 2 "modules", one of which is considered as a Z-Stage device. A stage widget will therefore be generated for these devices as well and I'm not sure if there will be any error (maybe when trying to move the stage through the control buttons...)

fdrgsp:
> There no longer seems to be a way to have the stage controls be part of the main controls layout - it's now triggered by a button

@ianhi  I think to does not matter much at the moment. I think that we can discuss about the gui "structure" once  all the widget will be ready (in a widget package). What do you think?

ianhi:
> @ianhi I think it does not matter much at the moment.

I've been using this to do experiments and definitely like this much less so it matters to me at least :). This is would be a decrease in usefulness to me and I would probably run a local branch re-integrating it instead of running off main. 

>  I believe we can discuss about the gui "structure" once all the widgets will be ready (in a widget package). What do you think?

Totally agree, but that's also why I think changes like this should not be included in PRs like this one. Improving the widget and changing the UI layout are really orthogonal to each other and it's confusing to lump them. So Instead of changing it piece by piece in PRs we should determine the overall plan and implement it wholesale and consistently.

ianhi:
Thanks for being flexible @fdrgsp - really appreciate it :)


I'm now noticing an issue wiht the QCollapsibles - which may be an issue from superqt (so heads up @tlambert03 )
(I'm running superqt HEAD)

1. Stages don't fully collapse
2. If you collapse the tab above then the stage section goes wild
Peek 2022-04-28 12-50

tlambert03:
yeah, I noticed that too.  it is in superqt.  and it came from the fix in https://github.com/napari/superqt/pull/72

 :/

ianhi:
That aside everything looks fine on the demo cam. I will try to try this out a real scope today or tomorrow (pending availability). Should we wait until that's fixed in superqt, or can we forge ahead assuming that superqt will be fixed before it's next release?

ianhi:
Ok! I tried this out on a real scope and have the following notes:


1. The ordering of stage widgets isn't the same between scopes. We should maybe consider first sorting by stage type, then by name. On the TE2000 I got XY last rather than first. This isn't a big issue, but it owuld be nice to be consistent between scopes

image


2. **important** There is no way to trigger the polling on the Z or autofocus widgets
     - For me the device drivers dont have callbacks, which means that these never update.

fdrgsp:
> 1. The ordering of stage widgets isn't the same between scopes.
At the moment the order depends on the device name (alphabetical order). But I made the widgets draggable, so you can manually change the position. Do you think we need to do something different?

> There is no way to trigger the polling on the Z or autofocus widgets
I will add the pool checkbox also to the z stages:)

ianhi:
>  But I made the widgets draggable,

oooh neat. We should try to make that more discoverable. I would never have figured that out if you hadn't told me.

ianhi (approved):
This is a good user experience and a nice improvement! Can we merge and the iterate more (autofocus etc) in future PRs?

ianhi:
> @ianhi I added a line of code and now we always display the XY stages first and then the Z stages (widgets still draggable).

:star_struck:

PR #144, opened 2022-03-09
Channel widget

fdrgsp:
merge after #164 

Add a `channel widget` based on `Generic PresetWidget` #128 and remove it from `main_window.py`.


micromanager_gui/_gui_objects/_channel_widget.py
```diff
@@ -0,0 +1,94 @@
...
+        self.channel_wdg.setParent(None)
+        self.channel_wdg.deleteLater()
+
+    def _update_widget(self, channel_group):
+        self.channel_wdg = self._create_channel_widget(channel_group)
+        self.layout().addWidget(self.channel_wdg)
+        if isinstance(
+            self.channel_wdg, PresetsWidget
+        ) and not self._mmc.getCurrentConfig(channel_group):
+            self.channel_wdg._combo.setStyleSheet("color: magenta;")
```
ianhi, line 90:
```suggestion
```

Just let the PresetWidget handle this. I think this is incorrectly causing it to be magenta on creation.

micromanager_gui/_gui_objects/_channel_widget.py
```diff
@@ -0,0 +1,94 @@
...
+    def _on_channel_group_changed(self, new_channel_group: str):
+        """When Channel group is changed, recreate combo."""
+        self._clear_previous_device_widget()
+        self._update_widget(new_channel_group)
+
+    def _clear_previous_device_widget(self):
+        self.channel_wdg.setParent(None)
+        self.channel_wdg.deleteLater()
+
+    def _update_widget(self, channel_group):
```
ianhi, line 84: Just combine these three methods?

ianhi (changes requested):
When you start it from the launch-dev script it seems to incorrectly mark the channel with magenta. otherwise looking pretty good!

micromanager_gui/_gui_objects/_channel_widget.py
fdrgsp: @ianhi you were right. To better handle that I made a small change in the PresetsWidget (#163).

tlambert03 (approved):
LGTM

Issue #143, opened 2022-03-09
MDA Image layer can overflow memory


ianhi:
## Problem

Currently there are no guards on the size of the MDA image layer:
https://github.com/tlambert03/napari-micromanager/blob/10a4b229fd3b0d1438ee3a2d750cb7d13ab3d2bf/micromanager_gui/main_window.py#L308-L313

given that MDAs can easily exceed the size of a standard computer's RAM I think this is setting us up for some crashes deep into an acquisition (probably by the time everyone has already gone home...).

## Solution

not sure what's best here. We could create a zarr file in a temp directory and display that - storing the images as they come in.

tlambert03:
We should also just have the option not to show the full dataset in the viewer (like most acquisition software).  We can also do some checks at the beginning of MDA for RAM, and hard drive space

PR #145, opened 2022-03-10
Back mda viewer with zarr to prevent memory overflow

ianhi:
Basic implementation to start. Uses `zarr` if it's available and falls back to `np.memmap` if not.

Not complete (or tested) but just wanted to check that this a direction we feel good about before diving in.

Limitations:
1. Doesn't account for channel splitting
    - will fix in this PR
2. Doesn't do the ram or disk space checks mentioned in #143 
3. No option to not show the whole dataset

I'd like to leave 2 and 3 out of this PR. They're important but I view this more as a bugfix than an expansion of features and would like to keep this as simple as possible so i can actually use this :)

closes: https://github.com/tlambert03/napari-micromanager/issues/143

thanks also to @jrussell25 for helpful discussions


micromanager_gui/main_window.py
```diff
@@ -271,11 +271,48 @@ def _update_mda_engine(self, newEngine: PMDAEngine, oldEngine: PMDAEngine):
         newEngine.events.sequenceStarted.connect(self._on_mda_started)
         newEngine.events.sequenceFinished.connect(self._on_mda_finished)
 
+    @ensure_main_thread
```
ianhi, line 274: This is not necessary on the demo camera on linux. but is on a real scope on windows. :shrug:

micromanager_gui/main_window.py
```diff
@@ -81,6 +84,21 @@ def __init__(self, viewer: napari.viewer.Viewer, remote=False):
         self._mmc.mda.events.sequenceStarted.connect(self._on_mda_started)
         self._mmc.mda.events.sequenceFinished.connect(self._on_mda_finished)
         self._mmc.events.mdaEngineRegistered.connect(self._update_mda_engine)
+        self._mda_temp_arrays = {}
+        self._mda_temp_files: Dict[str, tempfile.TemporaryDirectory] = {}
```
ianhi, line 88: Have to keep a reference to the tempdir otherwise it gets removed and our zarr/memmap will return only zeros.

ianhi: carries the real downside of now being in charge of cleaning up files though :(

ianhi:
There's a decent bit happening here so I've broken it up into two commits.

### 1.  https://github.com/tlambert03/napari-micromanager/pull/145/commits/fac952d1f23c7a9f780fba9091df3bf776fc3dfb

Implements the temporary files and moves all layer creation logic from `_on_mda_frame` to `_on_mda_started`

### 2. https://github.com/tlambert03/napari-micromanager/pull/145/commits/7345c8574bcc0e3fabe1d2154845be0ee4f7df48

makes everything work when the `Split Channels` option is checked. This adds a decent amount of complexity hence the separate commit for easier review (i hope...).

ianhi:
pre-commit failing because we need to update black as in: https://github.com/tlambert03/napari-micromanager/pull/160

ianhi:
ugh....

```
 ______________________ test_main_window_mda[local-numpy] ______________________
CALL ERROR: Exceptions caught in Qt event loop:
________________________________________________________________________________
  File "C:\hostedtoolcache\windows\Python\3.9.12\x64\lib\site-packages\superqt\utils\_ensure_thread.py", line 57, in _func
    return _run_in_thread(
  File "C:\hostedtoolcache\windows\Python\3.9.12\x64\lib\site-packages\superqt\utils\_ensure_thread.py", line 116, in _run_in_thread
    result = func(*args, **kwargs)
  File "D:\a\napari-micromanager\napari-micromanager\micromanager_gui\main_window.py", line 339, in _on_mda_started
    self._mda_temp_arrays[id_] = np.memmap(
  File "C:\hostedtoolcache\windows\Python\3.9.12\x64\lib\site-packages\numpy\core\memmap.py", line 228, in __new__
    f_ctx = open(os_fspath(filename), ('r' if mode == 'c' else mode)+'b')
PermissionError: [Errno 13] Permission denied: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmpe0cwlcsy'
```

apparently `TemporaryDirectory` works, but `NamedTemporaryFile` does not???

tlambert03:
looks like mac is still running, so this is probably a windows specific thing.  On windows, you need to close the file before you can touch it... try something like:

```python
tf = NamedTemporaryFile(delete=False)
tf.close()
try:
    ... # do something with tf.name
finally:
    os.unlink(tf.name)  # delete manually
```

ianhi:
> finally:
    os.unlink(tf.name)  # delete manually

The issue we can't do that in `on_mda_started` because we need the file to stick around during the whole mda. So I'm not sure that that will work?

ianhi:
> so this is probably a windows specific thing. 

I was thinking this is a windows on github actions specific thing

tlambert03:
> I was thinking this is a windows on github actions specific thing

nope, I've had to use that workaround for windows namedtempfile everywhere ...

> The issue we can't do that in on_mda_started because we need the file to stick around during the whole mda. So I'm not sure that that will work?

that's fine.  You don't need to clean it up immediately if you don't want to.  I assumed you were using NamedTemporaryFile as a context manager, that's why I gave that example... but if you aren't then you only need to call `tf.close()`

ianhi:
oooh i see thanks!

Alternatively solution is what I did here to just always use `TemporaryDirectory` which is think is better because it's consistent between zarr and numpy

ianhi:
I wonder if that failure is replicable. gonna restart them

ianhi:
This might be a bit complex to review - but it's also critical if this is going to be used for any long timelapses. If there's anything more I can do to make this easier to review let me know

launch-dev.py
```diff
@@ -11,8 +12,8 @@
 
 sequence = MDASequence(
     channels=["Cy5", {"config": "FITC", "exposure": 50}],
-    time_plan={"interval": 2, "loops": 5},
-    z_plan={"range": 4, "step": 0.5},
+    time_plan={"interval": 2, "loops": 2},
```
tlambert03, line 15: does this pr require this?

tlambert03 (commented):
I think one thing that would help me here is more comments, pulling stuff out into subfunctions with docstrings, type hints, etc...

I gather that we're creating a zarr store in a tempdir, passing that to napari with the shape/dtype of the _full_ experiment, then just updating the setpoint as each frame comes in.  that all sounds good to me, but I'd kind of expect to see helper functions like `create_store_for_sequence`, `write_to_store`, etc...  rather than having it all in the body of the `_on_mda_x`.

Could you break those out into smaller logical bits with names that explain more what they're doing?

micromanager_gui/main_window.py
```diff
@@ -14,14 +16,20 @@
 from qtpy.QtGui import QColor, QIcon
 from superqt.utils import create_worker, ensure_main_thread, signals_blocked
 
+try:
```
tlambert03, line 19: not necessary.  you added to requirements (which I like)

ianhi: oh woops, I think I intended to make it optional. But if you're happy with requiring zarr that will make it simpler because I can just eliminate the `memmap` approach and only use zarr.

ianhi:
> Could you break those out into smaller logical bits with names that explain more what they're doing?

good call. will do

launch-dev.py
ianhi: no but it saved me 6 seconds every time I would iterate. Will remove at the end

ianhi:
@tlambert03 I added comments and broke  `on_mda_started` up into functions. But I didn't see a sensible way to break up the `mda` branch of the if in `on_mda_frame` so I just left comments there.  Happy to break up further if you have a suggestion for how

ianhi:
I've also removed the `np.memmap` option, because if we are ok with requiring `zarr` then I don't see a good reason to add complexity by supporting both.

ianhi:
Latest commit uses a pytest fixture to get a temp dir in which to exervise the saving machinery. The only errors were related to that which this PR wasn't really touching. example error:

```
Exceptions caught in Qt event loop:
________________________________________________________________________________
  File "/Users/runner/work/napari-micromanager/napari-micromanager/micromanager_gui/main_window.py", line 476, in _on_mda_finished
    save_sequence(sequence, self.viewer.layers, meta)
  File "/Users/runner/work/napari-micromanager/napari-micromanager/micromanager_gui/_saving.py", line 26, in save_sequence
    return _save_mda_sequence(sequence, layers, meta)
  File "/Users/runner/work/napari-micromanager/napari-micromanager/micromanager_gui/_saving.py", line 35, in _save_mda_sequence
    folder_name = ensure_unique(path / file_name, extension="", ndigits=3)
  File "/Users/runner/work/napari-micromanager/napari-micromanager/micromanager_gui/_util.py", line 66, in ensure_unique
    for fn in paths:
  File "/Users/runner/work/napari-micromanager/napari-micromanager/micromanager_gui/_util.py", line 64, in <genexpr>
    else (f for f in p.parent.iterdir() if f.is_dir())
  File "/Users/runner/hostedtoolcache/Python/3.9.12/x64/lib/python3.9/pathlib.py", line 1160, in iterdir
    for name in self._accessor.listdir(self):
FileNotFoundError: [Errno 2] No such file or directory: '/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tmpn2v6vzm9'
```

ianhi:
I feel like im going crazy. The signal is emitted and it looks like it's within a second of starting

```
        with qtbot.waitSignal(mmc.mda.events.sequenceFinished, timeout=3000):
>           _mda._on_run_clicked()
E           pytestqt.exceptions.TimeoutError: Signal sequenceFinished(PyQt_PyObject) not emitted after 3000 ms
...
2022-05-09 21:42:13.220 | INFO     | pymmcore_plus.mda._engine:_prepare_to_run:123 - MDA Started: Multi-Dimensional Acquisition ▶ nt: 3, np: 1, nz: 3, nc: 2
....
2022-05-09 21:42:13.616 | INFO     | pymmcore_plus.mda._engine:_finish_run:235 - MDA Finished: Multi-Dimensional Acquisition ▶ nt: 3, np: 1, nz: 3, nc: 2
```

maybe the saving is taking a longer time. But that saving is something I ultimately want to remove from here.

Any ideas other than bumping that timeout a bit more? It's also weird that it only fails in one case

fdrgsp:
@Ian I don't think it is related but does the time-lapse settings work for you? When I setup a time-lapse, doesn't matter what interval I set, the acquisition does not wait for the interval...maybe there is a problem with the mda engine?

ianhi:
I'm hoping this will have been fixed by https://github.com/tlambert03/pymmcore-plus/pull/127 which fixed a bug wiht mda timing

ianhi:
hooorayyy I think it did! I will restart a few time to be sure

ianhi:
ok! I think this is consistently passing now!

micromanager_gui/main_window.py
```diff
@@ -219,68 +246,134 @@ def _update_mda_engine(self, newEngine: PMDAEngine, oldEngine: PMDAEngine):
...
+        img_shape = self._mmc.getImageWidth(), self._mmc.getImageHeight()
+        # dimensions labels
+        axis_order = event_indices(next(sequence.iter_events()))
+        labels = []
+        shape = []
+        for i, a in enumerate(axis_order):
+            dim = sequence.shape[i]
+            if dim != 1:
+                labels.append(a)
+                shape.append(dim)
```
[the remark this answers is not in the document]
ianhi: fixed

ianhi:
@tlambert03 I think this is now ready to go

tlambert03 (approved):
Thanks for all the work here @ianhi!  I've had a play and I'm happy to move forward with this.

I do think it's generally better and safer, and a step in the right direction.  but there are some subtle losses in performance with this approach for sure.  (I assume all the file IO is responsible).  In the future, it'd be nice to bring back the in-memory behavior as well if someone knows they won't be exceeding their RAM (along with checks to let you know whether you would be).  Fast time-lapses in particular will likely suffer a bit from this change.

I also think we should follow up soon to clear out the temporary directories when a layer gets deleted (and not at program close).

All in all though, a very nice option to have here!  LGTM and let's continue to work on the in-memory stuff too

PR #147, opened 2022-03-10
Snap button widget

fdrgsp:
Creating a new general purpose Snap `QPushButton` widget linked to the `pymmcore-plus` `snap` method.
Once the button is `clicked`, an image is snapped and the `pymmcore-plus` `imageSnapped` signal is emitted.


micromanager_gui/_core_widgets/_snap_button_widget.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def _create_button(self):
+        if self.button_text:
+            self.setText(self.button_text)
+        self.setIcon(icon(MDI6.camera_outline, color=self.icon_color))
+        self.setIconSize(QSize(self.icon_size, self.icon_size))
+        self.clicked.connect(self._snap)
+
+    def _snap(self):
+        loaded = self._mmc.getLoadedDevices()
```
tlambert03, line 54: rather than checking this on every snap, let's just try `_mmc.snap()` and catch an error.  Moreover... can we go one step further and listen for sys config loaded signals and disable the button whenever a camera isn't availalable?

ianhi: Also, if we do do a check like this (it could be moved to init) we should not just check for number of loaded devices but instead use either [`mmc.getLoadedDevicesOfType`](https://micro-manager.org/apidoc/MMCore/latest/class_c_m_m_core.html#a193eb1e3d80e3dabee70f1434a2b1b75) or '[`iter_devices`](https://pymmcore-plus.readthedocs.io/en/latest/api/pymmcore_plus.html#pymmcore_plus.CMMCorePlus.iterDevices)

micromanager_gui/_core_widgets/_snap_button_widget.py
```diff
@@ -0,0 +1,77 @@
...
+    ) -> None:
+
+        super().__init__()
+
+        self._mmc = mmcore or get_core_singleton()
+        self.button_text = button_text
+        self.icon_size = icon_size
+        self.icon_color = icon_color
+
+        self._mmc.events.systemConfigurationLoaded.connect(self._on_system_cfg_loaded)
```
ianhi, line 44:
```suggestion
        self._mmc.events.systemConfigurationLoaded.connect(self._on_system_cfg_loaded)
        self._on_system_cfg_loaded()
```

If you call it once so then it will be enabled if a config was loaded before button was created

fdrgsp:
I wanted to modify the snap method in the main_window like below:
```
sig.imageSnapped.connect(self._snap)

def _snap(self, img: np.ndarray):
        self.stop_live()
        create_worker(self.update_viewer, img, _start_thread=True)
```
but when the image is snapped and the `ndarray` is passed to the `update_viewer` function I get this error:
```
WARNING: QObject::connect: Cannot queue arguments of type 'QVector<int>'
(Make sure 'QVector<int>' is registered using qRegisterMetaType().)
```
Do you know why?

tlambert03:
Yeah that suggests your communicating with the gui across threads.  Updating the napari viewer should never happen in a non main thread.  You need to connect the viewer update to a callback after the worker is finished

micromanager_gui/_core_widgets/_snap_button_widget.py
```diff
@@ -0,0 +1,81 @@
...
+        self.clicked.connect(self._snap)
+
+    def _snap(self):
+        if self._mmc.isSequenceRunning():
+            self._mmc.stopSequenceAcquisition()
+
+        try:
+            self._mmc.snap()
+        except RuntimeError:
+            raise
```
ianhi, line 64: @fdrgsp I think the delegation to a thread needs to happen here in the widget rather than the main window. Otherwise this button will be capable of stalling the event loop of any gui it's in. Then the main window just needs the `update_viewer` (with ensure_main_thread) attached to the `imageSnapped` signal.

fdrgsp: Thanks @ianhi. Do you think is ok now?

micromanager_gui/main_window.py
```diff
@@ -220,16 +224,11 @@ def update_max_min(self, event=None):
 
         self.tab_wdg.max_min_val_label.setText(min_max_txt)
 
-    def snap(self):
+    def _snap(self, img: np.ndarray):
+        # _snap in a thread so we don't freeze UI
+        create_worker(self.update_viewer, img, _start_thread=True)
```
ianhi, line 229: As it stands this won't actually snap a new image - it just updates the viewer from a thread. I think what you want here is to create a thread worker to call snap just like in the button widget (sorry if I was confusing about that earlier)

micromanager_gui/_core_widgets/_snap_button_widget.py
ianhi: see other comment - was easier to talk about in th efull context

micromanager_gui/_core_widgets/_snap_button_widget.py
```diff
@@ -0,0 +1,84 @@
...
+
+        create_worker(self._mmc.snap, _start_thread=True)
```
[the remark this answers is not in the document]
fdrgsp: True, I forgot to remove that...

micromanager_gui/main_window.py
fdrgsp: True! But since the image is snapped through the `pymmcore-plus` `snap` method in `snap_button.py` and is "emitted" with the `pymmcore-plus` `imageSnapped` signal, here in main_window we just need to display it, correct?

ianhi: So `_snap` is used for the `snap_on_click` from the stage widget. So this needs to call snap for those. 

So I think there needs to be a  `_snap` calls `snap` from a thread, and then you can directly hook up `update_viewer` to the `imageSnapped` signal - that doesn't need to have a thread be created.

micromanager_gui/_core_widgets/_snap_button_widget.py
```diff
@@ -0,0 +1,82 @@
...
+        self.clicked.connect(self._snap)
+
+    def _snap(self):
+        if self._mmc.isSequenceRunning():
+            self._mmc.stopSequenceAcquisition()
+
+        try:
+            create_worker(self._mmc.snap, _start_thread=True)
+        except RuntimeError:
+            raise
```
ianhi, line 65: One more thing here - just realized that this try except gaurd won't actually catch the error as that error would occur inside the thread. Though it may not be necessary with good checks in init and the error in the thread probably won't ruin things.

micromanager_gui/main_window.py
```diff
@@ -220,16 +224,11 @@ def update_max_min(self, event=None):
 
         self.tab_wdg.max_min_val_label.setText(min_max_txt)
 
-    def snap(self):
+    def _snap(self, img: np.ndarray):
+        # update in a thread so we don't freeze UI
+        create_worker(self.update_viewer, img, _start_thread=True)
```
ianhi, line 229:
```suggestion
    def _snap(self, img: np.ndarray):
        self.stop_live()
        # run snap  in a thread so we don't freeze UI
        create_worker(
            self._mmc.snap,
            _start_thread=True,
        )
```

This can't just update the viewer - it also needs to trigger a new snap so that it works for `snap_on_click`

ianhi: Can probably remove this entirely in the future once the new stage widget comes in

ianhi:
@fdrgsp can you add some tests - even a super basic one like in https://github.com/tlambert03/napari-micromanager/pull/150?

After that I'm happy to give this a more thorough review

ianhi:
also, rebasing on main should fix the test failures

fdrgsp:
> @fdrgsp can you add some tests - even a super basic one like in #150?
> 
> After that I'm happy to give this a more thorough review

@ianhi basic test added!

micromanager_gui/main_window.py
fdrgsp: > So `_snap` is used for the `snap_on_click` from the stage widget. So this needs to call snap for those.

as I mention in #142, I love this "snap_after_stage_moves" function but I think we can remove it for now and see how to add it back once the stage, the snap and the live button widgets are merged. Do you agree?

tests/core_widgets/test_snap_button_widget.py
```diff
@@ -0,0 +1,25 @@
+from micromanager_gui._core_widgets._snap_button_widget import SnapButton
+
+
+def test_snap_button_widget(qtbot, global_mmcore):
```
ianhi, line 4: Add the typing now so it's nicer to develop on in the future.

```suggestion
from __future__ import annotations
from micromanager_gui._core_widgets._snap_button_widget import SnapButton

from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from pytestqt.qtbot import QtBot
    from pymmcore_plus import CMMCorePlus

def test_snap_button_widget(qtbot: QtBot, global_mmcore: CMMCorePlus):
```

tests/core_widgets/test_snap_button_widget.py
```diff
@@ -0,0 +1,25 @@
...
+    assert snap_btn.icon_size == 40
+    assert snap_btn.icon_color == "green"
+
+    global_mmcore.startContinuousSequenceAcquisition(0)
+
+    snap_btn.click()
+
+    assert not global_mmcore.isSequenceRunning()
+
+    with qtbot.waitSignal(global_mmcore.events.imageSnapped) as image:
```
ianhi, line 24: The `click` should happen inside this `with` block, otherwise the event may fire before we enter here and things will fail.

micromanager_gui/main_window.py
```diff
@@ -76,6 +76,8 @@ def __init__(self, viewer: napari.viewer.Viewer, remote=False):
         sig.stagePositionChanged.connect(self._on_stage_position_changed)
         sig.exposureChanged.connect(self._on_exp_change)
 
+        sig.imageSnapped.connect(self._snap)
```
ianhi, line 79: No need to update the viewer in a thread, just pass directly to the update function

```suggestion
        sig.imageSnapped.connect(self.update_viewer)
```

micromanager_gui/main_window.py
ianhi: @fdrgsp this still needs to be changed - currently snap on click is broken

ianhi (changes requested):
Looking good! But still need to fix snap on click.

The only final thing to consider would be to generalize the snap button widget to work for any camera, not just the default camera (e.g. see how #150 and #142 manage this). But this is a good PR otherwise and we can always expand that later.

fdrgsp:
test error due to `codecov/patch — 74.07% of diff hit (target 80.00%)`

micromanager_gui/main_window.py
```diff
@@ -215,10 +219,9 @@ def update_max_min(self, event=None):
 
         self.tab_wdg.max_min_val_label.setText(min_max_txt)
 
-    def snap(self):
+    def _snap(self):
         self.stop_live()
-
-        # snap in a thread so we don't freeze UI when using process local mmc
+        # update in a thread so we don't freeze UI
```
ianhi, line 224: I think we should simplifying the create worker here to just be:

```python
create_worker(self._mmc.snap, _start_thread=True)
```
and let the callbacks handle updating the viewer. That will allow other sources listening to the imageSnapped signal to pick it up.

ianhi (approved):
One comment on a small improvment - but on the whole this looks good

micromanager_gui/main_window.py
```diff
@@ -524,41 +523,41 @@ def stage_x_left(self):
...
-            self.snap()
+            self._snap()  # pragma: no cover
 
     def stage_z_down(self):
         self._mmc.setRelativePosition(
             -float(self.stage_wdg.z_step_size_doubleSpinBox.value())
         )
         if self.stage_wdg.snap_on_click_checkBox.isChecked():
-            self.snap()
+            self._snap()  # pragma: no cover
```
ianhi, line 563: I don't think these `no cover`s are good to include. This is legitimately untested and is something that should be tested. Though that can actually be implemented in #142

fdrgsp: @ianhi  I agree with you! I did add `no covers` here because will be gone in 142 (I'm now working on this PR too).

ianhi:
@tlambert03 I think this is ready and would be a nice to have as a separate widget for the demo. Any chance of merging this todayish?

tlambert03 (approved):
good stuff.  Thanks!

Issue #127, opened 2022-03-03
Fixes for live mode


ianhi:
We can hold off on this for a bit till happy with #111 but I suspect we have some bugs in live mode.

See extensive explanation here: https://github.com/micro-manager/mmCoreAndDevices/issues/171#issuecomment-1058441397

In particular we need to:
1. Be even more careful than we currently are when updating the exposure
2. Not bother with passing a number to startContinuousSequenceAcquistion because it doesn't actually affect anything.

AlexCoul:
It's true, I was about to post an issue about that but I noticed this post, and I actually found a quick fix:
When I tried to use the Live mode, I had the error
```
TypeError                                 Traceback (most recent call last)
File ...\napari-micromanager\micromanager_gui\main_window.py:217, in MainWindow._start_live(self=<micromanager_gui.main_window.MainWindow object>)
    215 self.streaming_timer = QTimer()
    216 self.streaming_timer.timeout.connect(self.update_viewer)
--> 217 self.streaming_timer.start(self._mmc.getExposure())
        self.streaming_timer = None
        self = <micromanager_gui.main_window.MainWindow object at 0x000001372664B910>
        self._mmc = <CMMCorePlus at 0x13726666de0>

TypeError: arguments did not match any overloaded call:
  start(self, int): argument 1 has unexpected type 'float'
  start(self): too many arguments
```
so in `main_window.py` line 217 I changed `self.streaming_timer.start(self._mmc.getExposure())` to `self.streaming_timer.start(int(self._mmc.getExposure())))`, and now the Live mode works :-)

ianhi:
@AlexCoul would you like to open a PR with your fix?

AlexCoul:
Hi @ianhi, yep I just did!
As I said it's super small, but on my machine I can now use the live mode.

fdrgsp:
fixed in #181

PR #148, opened 2022-03-11
Live mode button widget

fdrgsp:
Creating a new general purpose two-state (on-off) `Live` mode `QPushButton` widget.

When toggled on,  a signal is emitted (`_emitFrameSignal`) through a `QTimer`. In `main_window` the signal is connected to the `update_viewer` function.

@tlambert03 @ianhi Do you think it can be a valid solution?

### DIFF
--- a/micromanager_gui/_core_widgets/_live_button.py
+++ b/micromanager_gui/_core_widgets/_live_button.py
@@ -1,11 +1,14 @@
 from typing import Optional, Tuple, Union
 
 from fonticon_mdi6 import MDI6
+
+# from numpy import ndarray
 from pymmcore_plus import CMMCorePlus
-from qtpy.QtCore import QSize, Qt
+from qtpy.QtCore import QSize, Qt, QTimer, Signal
 from qtpy.QtGui import QColor
 from qtpy.QtWidgets import QApplication, QPushButton
 from superqt.fonticon import icon
+from superqt.utils import create_worker
 
 from micromanager_gui._core import get_core_singleton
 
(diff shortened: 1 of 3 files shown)


ianhi:
> When toggled on, a signal is emitted (_emitFrameSignal) through a QTimer. In main_window the signal is connected to the update_viewer function.
> 
> @tlambert03 @ianhi Do you think it can be a valid solution?

This is a really good start! But I think we can go farther in the decoupling if we can stomach adding a new core signal. I think my ideal would be:

1. core gets the new signals: `continuousSequenceAcquisition[Started/Stopped]`
2. Live button is only responsible for:
     - stopping and starting the sequence
     - Temporarily pausing the acquisition when the exposure is updated
         (actually this could maybe even be moved onto core)
3. main_window listens to the new signals and starts its own QTimer based on the exposure


This works out if all listeners call `getLastImage` because that doesn't actually remove anything from the circular buffer. 

I think that that's more or less analogous to what your doing with the snap button. This is also y preference because it feels like a cleaner separation of concerns. The live button is just responsible for toggling a single state on core - not managing timers and emitting data. Ideally the only place other components receive data from is the core.


Alternatively, we can push the timer logic into the button as you've done now. But in that case I think we should:
1. make the signal public
2. consider changing the name  - in this case it's really more of a `live_manager` than a button.

micromanager_gui/_core_widgets/_live_button.py
```diff
@@ -0,0 +1,116 @@
...
+                self.setText(self.button_text_on)
+
+        else:
+            self.setIcon(icon(MDI6.video_off_outline, color=self.icon_color_off))
+            if self.button_text_off:
+                self.setText(self.button_text_off)
+        self.setIconSize(QSize(self.icon_size, self.icon_size))
+
+    def start_live(self):
+        self._mmc.startContinuousSequenceAcquisition(self._mmc.getExposure())
```
ianhi, line 89:
```suggestion
        self._mmc.startContinuousSequenceAcquisition(0)
```

Argument doesn't have any effect see https://github.com/tlambert03/napari-micromanager/issues/127. I think copying micro-manager and using 0 here is probably best.

fdrgsp:
some tests are failing because of the new  signals not yet added to `pymmcore-plus`.

tlambert03 (approved):
This looks good to me, couple minor comments on the tests.  @ianhi, I think this generally follows the pattern you were also endorsing.  Good with this?

ianhi:
> This looks good to me, couple minor comments on the tests. @ianhi, I think this generally follows the pattern you were also endorsing. Good with this?


missed this but yes it does! nice stuff

PR #158, opened 2022-03-18
update opengl git url

ianhi:
see https://stackoverflow.com/questions/70663523/the-unauthenticated-git-protocol-on-port-9418-is-no-longer-supported

should hopefully fix the failing windows tests

Example failure from https://github.com/tlambert03/napari-micromanager/runs/5592092388?check_suite_focus=true

image


ianhi:
Looks like that worked.

@tlambert03 if you have a moment this should be an easy merge and make reviewing other PRs much easier on the eyes :)

PR #159, opened 2022-03-18
cleanup utils

ianhi:
two of these were non-functional and `blockSignals` should be superceded by the version in superqt


tlambert03 (approved):
love deletions :)

PR #164, opened 2022-04-22
minor update _presets_widget.py

fdrgsp:
Set the group combobox to the current preset when the widget is created or refreshed.

This is important for PR #144  and #122


micromanager_gui/_core_widgets/_presets_widget.py
```diff
@@ -125,6 +127,8 @@ def refresh(self) -> None:
                 presets = self._mmc.getAvailableConfigs(self._group)
                 self._combo.addItems(presets)
                 self._combo.setEnabled(True)
+                self._combo.setCurrentText(self._mmc.getCurrentConfig(self._group))
+                self._set_if_props_match_preset()
```
ianhi, line 131: I think you can just do this?

```suggestion
                self._set_if_props_match_preset()
```

ianhi: @fdrgsp bump here. I was looking at the channel widget PR and arrived back here.

micromanager_gui/_core_widgets/_presets_widget.py
```diff
@@ -38,6 +38,8 @@ def __init__(
 
         self._combo = QComboBox()
         self._combo.addItems(self._presets)
+        self._combo.setCurrentText(self._mmc.getCurrentConfig(self._group))
+        self._set_if_props_match_preset()
```
ianhi, line 42: same here i think
```suggestion
        self._set_if_props_match_preset()
```

tlambert03:
looks like this got added also in #167, so will close

PR #165, opened 2022-04-25
minor explorer fix/update

fdrgsp:
- `self._mmc.cancel()` -> `self._mmc._mda_engine.cancel()`
- `self.viewer.layers.remove("preview")` on explorer start

### DIFF
--- a/micromanager_gui/_gui_objects/_sample_explorer_widget.py
+++ b/micromanager_gui/_gui_objects/_sample_explorer_widget.py
@@ -62,7 +62,7 @@ class ExploreSample(QtW.QWidget):
         self.move_to_Button.clicked.connect(self.move_to)
         self.browse_save_explorer_Button.clicked.connect(self.set_explorer_dir)
 
-        self.stop_scan_Button.clicked.connect(lambda e: self._mmc.cancel())
+        self.stop_scan_Button.clicked.connect(lambda e: self._mmc._mda_engine.cancel())
 
         self.x_lineEdit.setText(str(None))
         self.y_lineEdit.setText(str(None))
--- a/micromanager_gui/main_window.py
+++ b/micromanager_gui/main_window.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import contextlib
 from collections import defaultdict
 from pathlib import Path
 from typing import TYPE_CHECKING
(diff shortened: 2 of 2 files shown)


micromanager_gui/_gui_objects/_sample_explorer_widget.py
```diff
@@ -62,7 +62,7 @@ def __init__(self, parent=None):
         self.move_to_Button.clicked.connect(self.move_to)
         self.browse_save_explorer_Button.clicked.connect(self.set_explorer_dir)
 
-        self.stop_scan_Button.clicked.connect(lambda e: self._mmc.cancel())
+        self.stop_scan_Button.clicked.connect(lambda e: self._mmc._mda_engine.cancel())
```
ianhi, line 65: This should use the public API in case we change the pymmcore-plus internals later.

```suggestion
        self.stop_scan_Button.clicked.connect(lambda e: self._mmc.mda.cancel())
```

ianhi (changes requested):

micromanager_gui/main_window.py
```diff
@@ -275,6 +276,10 @@ def _on_mda_started(self, sequence: useq.MDASequence):
             # originated from user script - assume it's an mda
             self._mda_meta.mode = "mda"
 
+        if self._mda_meta.mode == "explorer":
+            with contextlib.suppress(ValueError):
+                self.viewer.layers.remove("preview")
+
```
ianhi, line 282: Why do this?

ianhi: What happens if you do explorer and then start live mode - does the same problem crop up again?

fdrgsp: The microscope is not free at the moment, I need to check this better probably...I will record a video and show you what I mean. My point is that since napari has a nice world coordinates system, I'd love to use it to map the microscope stage...

ianhi: > My point is that since napari has a nice world coordinates system, I'd love to use it to map the microscope stage...

Absolutely! I just didn't understnad the issue at first. That sounds maybe complicated to get right - perhaps best to open an issue dedicated tothis?

> Anyway, I reverted the last change, let's leave this PR only for the mda.cancel() method

that keeps this nice and simple. This is good to go from me

PR #167, opened 2022-05-03
add ToolTip to any preset widget

fdrgsp:
I think it can be useful to know all the `properties` that have been assigned to a `group/preset` without using a "future" edit button/menu (we don't have it yet but I want to add it at some point) like in micromanager. 

This PR just adds `ToolTip` info depending on the `group-preset` `mmcore` `getConfigData()`.

It will work as in the examples below:

Screen Shot 2022-05-02 at 11 47 42 PM

Screen Shot 2022-05-03 at 9 19 45 AM

Screen Shot 2022-05-03 at 9 20 30 AM

### DIFF
--- a/micromanager_gui/_core_widgets/_presets_widget.py
+++ b/micromanager_gui/_core_widgets/_presets_widget.py
@@ -37,9 +37,9 @@ class PresetsWidget(QWidget):
         self._check_if_presets_have_same_props()
 
         self._combo = QComboBox()
+        self._combo.currentTextChanged.connect(self._update_tooltip)
         self._combo.addItems(self._presets)
         self._combo.setCurrentText(self._mmc.getCurrentConfig(self._group))
-        self._set_cfg_data_as_tool_tip(self._combo.currentText())
         self._set_if_props_match_preset()
 
         self.setLayout(QHBoxLayout())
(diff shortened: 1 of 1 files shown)


micromanager_gui/_core_widgets/_presets_widget.py
```diff
@@ -100,6 +105,7 @@ def _on_cfg_set(self, group: str, preset: str) -> None:
             with signals_blocked(self._combo):
                 self._combo.setCurrentText(preset)
                 self._combo.setStyleSheet("")
+                self._set_cfg_data_as_tool_tip(preset)
```
tlambert03, line 108: it seems like you're manually calling `_set_cfg_data_as_tool_tip` whenever the text of `self._combo` changes... so perhaps you should just connect to the `currentTextChanged` signal?

```
self._combo.currentTextChanged.connect(self._update_tooltip)

def _update_tooltip(self, preset):
    self._combo.setToolTip(str(self._mmc.getConfigData(self._group, preset)) if preset else '')
```

micromanager_gui/_core_widgets/_presets_widget.py
```diff
@@ -38,6 +38,9 @@ def __init__(
 
         self._combo = QComboBox()
         self._combo.addItems(self._presets)
+        self._combo.setCurrentText(self._mmc.getCurrentConfig(self._group))
+        self._set_cfg_data_as_tool_tip(self._combo.currentText())
+        self._set_if_props_match_preset()
```
tlambert03, line 43: same here... could this be connected to a change signal rather than called manually in "all the right places"?

tlambert03 (approved):

PR #168, opened 2022-05-05
Replace Explorer .ui gui file with .py

fdrgsp:
I recreated the `sample_explorer` gui widget in python and replaced the old .ui file.


tlambert03 (approved):
LGTM

PR #169, opened 2022-05-05
refactor: Replace MDA .ui gui file with .py

fdrgsp:
I recreated the `mda` gui widget in python and replaced the old .ui file.


tlambert03 (approved):
I'm going to optimistically merge this one.  I _think_ it has the side-effect of changing a couple of the default settings for MDA, but they were somewhat arbitrarily chosen to begin with, and I don't think the new defaults here make any less sense than the old ones.

I propose we have a follow up PR that lets the user somehow set their preferred startup defaults (based, e.g. on a uses-schema YAML or something)

PR #173, opened 2022-05-19
use build instead of setup.py

ianhi:
Release failed due to this https://github.com/tlambert03/napari-micromanager/runs/6515242667?check_suite_focus=true

attn: @tlambert03

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -75,9 +75,9 @@ jobs:
         run: |
           git tag
           pip install --upgrade pip
-          pip install -U setuptools setuptools_scm wheel twine check-manifest
+          pip install -U setuptools twine check-manifest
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -75,9 +75,9 @@ jobs:
         run: |
           git tag
           pip install --upgrade pip
-          pip install -U setuptools setuptools_scm wheel twine check-manifest
+          pip install -U setuptools twine check-manifest
```
tlambert03, line 78:
```suggestion
pip install -U build twine check-manifest
```

thanks,  we need `build` though ... not setuptools (build will create an env and grab setuptools based on pyproject.toml)

tlambert03 (approved):

PR #181, opened 2022-06-30
fix: Fix Live mode (#127)

AlexCoul:
Small PR to solve broken live mode.


tlambert03:
thanks!  works for me.  May I ask what Qt backend and version you're on?

AlexCoul:
Sure, I'm using PyQt5 5.12.3 with QtPy  2.1.0

PR #185, opened 2022-07-07
fix: always track the current camera for live and snap buttons

ianhi:
Previously if you accidently loaded the demo config before loading a real one then the `snap` button would fail because `self._camera` was not being updated. This does remove the ability to pass in a user chosen camera, but I don't think that that's very useful because both `snap` and `startContinuousSequenceAcquisition` don't actually take camera arguments.


tlambert03 (approved):
looks good to me.  @fdrgsp, please note this in the widgets breakout you're creating (actually .. the sooner we can use that here the easier it will be to avoid desynch)

PR #188, opened 2022-07-11
refactor: use pymmcore-widgets

tlambert03:


tlambert03:
ok, the main reason this is failing is that it's ending up with a different core instance.  So now, we get to really experience the joys of singleton management :joy:

it would be relatively easy to fix by not using a custom session core in our tests here, but I don't really want to do that yet.  (it's _nice_ to be able to use an isolated little core instance, particularly for tests).  So we may have to either create a nice `set_core` context manager, that allows you to inject a given core instance (we could potentially use [`in-n-out`](https://github.com/tlambert03/in-n-out) here), or, we might need to laboriously pass the core instance around through all the widgets (ew)

tlambert03:
see #194 instead

PR #191, opened 2022-07-12
add linux to gh-action test matrix

RileyMShea:
# Linux Support PR

This PR adds (ubuntu)linux to the test matrix.  

[Output of a gh-action full matrix of working runs](https://github.com/RileyMShea/napari-micromanager/actions/runs/2658128250)

## References

Much of this was adapted from:
- [the micro-manager gh-action workflow](https://github.com/micro-manager/micro-manager/blob/main/.github/workflows/apidoc.yml).  
- [pytest-qt docs on gh-actions](https://pytest-qt.readthedocs.io/en/latest/troubleshooting.html#github-actions)
- [mmcore install notes](https://github.com/micro-manager/mmCoreAndDevices#building-on-mac-and--linux)

## Summary

- Replaced powershell `pyvista` step with pyvista's cross-platform gh-action for headless `DISPLAY` setup.
- added linux step for apt-install build prereqs.
- added linux steps for swig-3 build, install and caching.
- added linux steps for mmcore build, install and caching.


.github/workflows/test.yml
```diff
@@ -51,6 +50,67 @@ jobs:
         # need to figure out what happened on mac with the utilities
         run: python -m pymmcore_plus.install -r 20220420
 
+      - name: Install Linux Micro-Manager build tools and dependencies
```
tlambert03, line 53: let's split all this out into a separate `test_linux` test, then we can do away with the `if: runner.os` in most of the steps... Also we don't need to parametrize quite so much (particularly given how much slower it is than the other tests), so let's just test `ubuntu-latest` with python 3.10 on pyqt5 (or, if you'd like to parametrize either the backend or the python version x2 that's fine)

.github/workflows/test.yml
```diff
@@ -51,6 +50,67 @@ jobs:
...
+
+      - name: Cache SWIG-3
+        if: runner.os == 'Linux'
+        id: cache-swig-3
+        uses: actions/cache@v2
+        with:
+          path: swig-3.0.12
+          key: mmstudio-javadoc-swig-3.0.12
+
+      - name: Build SWIG-3
```
tlambert03, line 73: i know this is just copied from micro-manager, and this can be played with in a follow up PR, but the linux runner actually [comes with swig 4.0.2-1ubuntu1](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2204-Readme.md).  Do you happen to know if swig 3? is required?

tlambert03 (commented):
thanks so much!  Excellent work!

Couple minor requests (split it into a linux test, don't parametrize, and try to use xvfb-action instead)... but then LGTM!  Thanks for taking the time to do this

.github/workflows/test.yml
```diff
@@ -51,6 +50,67 @@ jobs:
...
+        if: runner.os == 'Linux'
+        run: |
+          sudo apt-get update
+          sudo apt-get -y install build-essential autoconf automake libtool autoconf-archive pkg-config
+          sudo apt-get -y install libboost-all-dev
+          sudo apt-get -y install libpcre3-dev  # For SWIG-3
+          sudo apt install -y qt5-default libgl1-mesa-glx xvfb # Qt(5) and headless display dependencies
+          # Below are specific to gh actions https://pytest-qt.readthedocs.io/en/latest/troubleshooting.html#github-actions
+          sudo apt install libxkbcommon-x11-0 libxcb-icccm4 libxcb-image0 libxcb-keysyms1 libxcb-randr0 libxcb-render-util0 libxcb-xinerama0 libxcb-xfixes0
+          /sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -screen 0 1920x1200x24 -ac +extension GLX
```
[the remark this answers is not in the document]
tlambert03: ah, one more thing... for the qt libs on linux, add 

```yaml
steps:
- uses: tlambert03/setup-qt-libs@v1
```

.github/workflows/test.yml
tlambert03: annoying :joy:

ianhi: Also it's likely faster to just use conda to install swig-3  that's what I settled on for the build instructions for micromanager

PR #214, opened 2022-11-21
[WIP] feat: new GUI with all pymmcore-widgets

fdrgsp:
I know this might be a too big PR to merge/review and I can try to split it, however it would be great to have your opinion first @tlambert03 @ianhi! 

The PR shows a possible new updated GUI for `napari-micromanager`.

Compare to the previous versions, here I used `qtpy` `QToolBar`. Some `pymmcore-widgets` have been added directly, other can be opened through some `QPushButton` with icons. **Every** widget is a `QDockWidget` so users can dock them to `napari` if they want. MDA and Sample Explorer are also both `QDockWidget` and can be opened with two dedicated  `QPushButton`.

https://user-images.githubusercontent.com/70725613/202931891-06d2347d-c1e3-4484-b595-e5d52c0a60ae.mov

---
I've also added few nice features including:
 - the _scale_ to the layers we acquire so if they are z-stacks we can visualize them in 3D in napari.
 - if you right-click with the mouse on any image/stack acquired, you can move the xy and/or z stage to the position clicked or you can add the position to the MDA or Explorer.

https://user-images.githubusercontent.com/70725613/202946194-7eb8c28b-bda1-4192-8c69-0cbc36b3e2b6.mov

PR #217, opened 2022-11-29
feat: add widgets as QtViewerDockWidget

fdrgsp:
PR to break up from #214.

This PR shows how I think we can  add the `pymmcore-widgtes` we are using in `napari-micromanager` as `QtViewerDockWidget`.

Currently, the new `QtViewerDockWidget` are added to the "Micro-Manager" menu in the napari menu bar.

In the next PR, the menu bar will be replaced by a `Qtoolbar` and the `QtViewerDockWidget` will be called by `QPushButtons`.

---
Screen Shot 2022-11-29 at 10 02 31 PM


tlambert03:
it's still unclear (without looking through the code carefully) exactly what you're trying to do here from the description above... from the description, this _seems_ like a PR that would change a mechanism for adding widgets to the main menu... but I see at least 7 new widget files added too?

Also, it looks like you've added a bunch of stuff to the menu, with a note that it will all be removed in the next PR.  The point of incremental PRs is not to add a bunch of stuff as an example, only to remove it in a follow up PR... it's to add a small unit of easily reviewable code *even if it's not hooked up anywhere*... That is, you don't necessarily need to offer a way to access _anything_ in the GUI in this or any other PR.  So, it's still not totally clear to me what the _key_ thing added here is.  Is it all the new widgets?   Is it `_add_dock_widget`? will `_show_dock_widget` remain in follow ups, or is it just a "demo"?  and why should _update_min_max need to be involved here?   

can you clarify what the point of the PR is?

fdrgsp:
> can you clarify what the point of the PR is?

So with this PR I mainly wanted to convert to `QtViewerDockWidget` most of the `pymmcore-widgtes` we are using in `napari-micromanager`. 

The reason why you see '_7 new widget files_' is because I think we need the pymmcore-widgtes` to be `QDialogs` to be added as `QtViewerDockWidget` and I had to create a new file for them.

Once `QtViewerDockWidget`, I could not add them all to the GUI (as they were before) because they are too many and so I thought that it was ok to add the widgets to the menu since it was already there.

both the new `_add_dock_widget` and `_show_dock_widget` will remain in future PR, the only difference will be that the `QtViewerDockWidget` will be called form `QPushButtons` in a `QToolBar` instead from the menu bar.


Is it better if I do one widget at a time maybe? Like a PR only to convert the camera widget as `QtViewerDockWidget`, etc?

tlambert03:
>both the new _add_dock_widget and _show_dock_widget will remain in future PR, the only difference will be that the QtViewerDockWidget will be called form QPushButtons in a QToolBar instead from the menu bar.

Thanks!  that's good info to include.  You can remove the stuff in add_menu then.  It's sufficient to have to programmatically add them.

> Is it better if I do one widget at a time maybe? Like a PR only to convert the camera widget as QtViewerDockWidget, etc?

no, it's fine to do them all at once. but the explanation ("I think we need the pymmcore-widgets to be QDialogs to be added as `QtViewerDockWidget` and I had to create a new file for them.") is definitely helpful... I'm not sure why they'd need to be qdialogs?  (Dialogs are usually what you need if you want them to be their own individual window, with the option of an `exec_` method)

----

adding tests is also a good way to indicate: "here is the new functionality being added here" :)

micromanager_gui/main_window.py
```diff
@@ -110,36 +126,128 @@ def cleanup():
...
+        """Method to show the selected QtViewerDockWidget."""
+        # TODO: temporary suolution, this will be
+        # changed in the next PR where the dockwidgets
+        # will be called by QPushButtons in a Qtoolbar
+
+        text = self.sender().text()
+
+        if "Browser" in text:
+            if not hasattr(self, "prop_browser_dock_wdg"):
+                self.prop_browser_dock_wdg = self._add_dock_widget(
```
tlambert03, line 187: not a big fan of this general `if not hasattr` pattern.  It makes it very hard to see (in the `__init__` for example) what this `MainWindow` class *might* have.  moreover, all of these are added as public attributes, making it very tempting for someone to access them... which is almost certain to lead to an `AttributeError` given that they are lazily added

tlambert03: also, the matching up of strings from the action names to the dock widget that gets added is too procedural.  I would expect to see a mapping somewhere of all the valid keys and their corresponding widgets:

```python
        WIDGETS = {
            "Device Property Browser": self.prop_browser,
            "Groups and Presets": self.group_preset_table_wdg,
            "Illumination Control": self.illumination,
            "Stages Control": self.stages,
        }
```

after which `_add_menus` could have just been:
```python
for name in WIDGETS:
    action = self._menu.addAction("Groups and Presets")
    action.triggered.connect(self._show_dock_widget)
```

and all of `_show_dock_widget` could just be:

```python
text = self.sender().text()
wdg = self._add_dock_widget(WIDGETS[text], text, floating=True)
wdg.show()
wdg.raise_()
```

(and use a dictionary somewhere to remember what widgets already have dock widgets

fdrgsp: what about adding them at strtup but hidden?

```
def __init__(self, viewer: napari.viewer.Viewer, remote=False):
        super().__init__()
        ...
        self.camera_roi_dock_wdg = self._add_dock_widget(
                   self.cam_roi, "Camera ROI", floating=True, hide=True
        )
        ...
```
and then 
```
def _show_dock_widget(self) -> None:
       """Method to show the selected QtViewerDockWidget."""
       text = self.sender().text()

       if "Camera" in text:
           self.camera_roi_dock_wdg.show()
           self.camera_roi_dock_wdg.raise_()
```

tlambert03: maybe ... but should check whether there's any noticeable difference in load time.  And I would still probably use a dictionary rather than attributes, since it's much easier to extend as you add more things

fdrgsp:
> I'm not sure why they'd need to be qdialogs? 

I know, and if I create and hide them at startup, everything is working well.

Instead, if I use this pattern (that anyway we are removing :)):

```
if "Camera" in text:
         if not hasattr(self, "camera_roi_dock_wdg"):
             self.camera_roi_dock_wdg = self._add_dock_widget(
                 self.cam_roi, "Camera ROI", floating=True
             )
         wdg = self.camera_roi_dock_wdg
         wdg.show()
         wdg.raise_()
```

there is a weird thing in the GUI at startup (top left of the image below):


Screen Shot 2022-12-01 at 8 33 08 PM

tlambert03:
ok, let's have a brainstorm together on this one.  maybe tomorrow?

micromanager_gui/main_window.py
fdrgsp: something like that?
```
def __init__(self, viewer: napari.viewer.Viewer, remote=False):
        super().__init__()
        ...
        self.camera_roi_dock_wdg = self._add_dock_widget(
                   self.cam_roi, "Camera ROI", floating=True, hide=True
        )
  
        self.dock_widgets = {
            "Camera ROI": self.camera_roi_dock_wdg,
            ...
            ...
        }
        ...
```
and
```
def _show_dock_widget(self) -> None:
        """Method to show the selected QtViewerDockWidget."""
        wdg = self.dock_widgets[self.sender().text()]
        wdg.show()
        wdg.raise_()
```

fdrgsp: ok I did't see the comment above, sorry

fdrgsp:
> ok, let's have a brainstorm together on this one. maybe tomorrow?

Sure!!!

tlambert03:
nope... I just reran the tests here and they're failing again.  @fdrgsp ... can you have a look?  I'm not sure why but we're getting a lot of sporadic failures now in `test_saving_explorer[withTranslate`

```
with qtbot.waitSignal(mmc.mda.events.sequenceFinished, timeout=4000):
[130](https://github.com/pymmcore-plus/napari-micromanager/actions/runs/3605462987/jobs/6081887096#step:7:131)
>           _exp.buttons_wdg.run_button.click()
[131](https://github.com/pymmcore-plus/napari-micromanager/actions/runs/3605462987/jobs/6081887096#step:7:132)
E           pytestqt.exceptions.TimeoutError: Signal  not emitted after 4000 ms
```

tlambert03:
you might just need to increase the timeout for github... (but let's also take a look to see whether we can speed up those tests)

tlambert03:
yeah, I think that was it... merged in #218 with the fix

tlambert03:
ok, merging.  you can implement the toolbar in the next PR.  then it's big cleanup time :)

PR #219, opened 2022-12-03
build: rename top module to `napari_micromanager`, use hatch and move to pyproject.toml only

tlambert03:
as the title sounds.

This renames the top module from `micromanager_gui` to `napari_micromanager` ... since we haven't even hit 0.1.0 and have no non-rc releases, I think it's fine to do without deprecation.


tlambert03:
@fdrgsp ... would you prefer that I wait on this one until after your next widget PR?  or ok to have it in whenever

fdrgsp:
> @fdrgsp ... would you prefer that I wait on this one until after your next widget PR? or ok to have it in whenever

Go for it!

PR #221, opened 2022-12-04
feat: replace menu with QToolBar 

fdrgsp:
This PR replaces the menu with a `QToolBar`.

Some of the `pymmcore-widgets` are now directly added to the `QToolBar`, other can be called with `QPushButtons` added to the toolbar.

The creation of the `QToolBar` and **all** the `QPushButtons`/`pymmcore-widgets` connections have been moved from `main_window.py` to a single file called `_toolbar.py`.

---
Screen Shot 2022-12-04 at 12 08 08 AM


src/napari_micromanager/_gui_objects/_toolbar.py
```diff
@@ -0,0 +1,368 @@
+from __future__ import annotations
+
+from pathlib import Path
+from typing import TYPE_CHECKING, Dict, Tuple, cast
+
+from fonticon_mdi6 import MDI6
+from napari._qt.widgets.qt_viewer_dock_widget import QtViewerDockWidget
```
tlambert03, line 7: importing from a private module is always risky.  In this case, since you only need it for a type hint, you should _definitely_ put it inside of `if TYPE_CHECKING`

tlambert03: even better ... just use `QtWidgets.QDockWidget` for the type hint.  Notice that outside of `_add_dock_widget`, the only part of the dockwidget api that you use is `show` and `raise` ... so from a typing perspective, you're not using the full `QtViewerDockWidget` object api.  The one exception is the line `dock_wdg._close_btn = False`, but that happens inside the scope of `_add_dock_widget`, so the type checker doesn't care

src/napari_micromanager/_gui_objects/_toolbar.py
```diff
@@ -0,0 +1,368 @@
...
+        """Add a docked widget using napari's add_dock_widget."""
+        dock_wdg = self.viewer.window.add_dock_widget(
+            widget,
+            name=name,
+            area="right",
+            allowed_areas=["left", "right"],
+            tabify=tabify,
+        )
+        dock_wdg.setFloating(floating)
+        dock_wdg._close_btn = False
```
tlambert03, line 367:
```suggestion
        try:
            dock_wdg._close_btn = False
        except AttributeError:
             pass
```

anytime you access a private attribute of an external dependency like this, you should assume it will disappear someday and try to make your code still work

tlambert03:
@fdrgsp ... it seems like you saw https://github.com/fdrgsp/napari-micromanager/pull/33 ... but didn't merge it.  how come?

tlambert03:
can you change all of those strings to `MDI6.___` and remove the getattr call, as in my PR?

fdrgsp:
> @fdrgsp ... it seems like you saw [fdrgsp#33](https://github.com/fdrgsp/napari-micromanager/pull/33) ... but didn't merge it. how come?

I looked at it on the phone and deleted the email and I didn't find it anymore :joy: I thought you added it in this PR and did't realize it was on mine, sorry

fdrgsp:
@tlambert03 do you like that when we add the widgets they are already docked? What about docking only the mda and explorer and leave the other undocked when created?

tlambert03:
yeah, I do like it, for everything but the property browser... but lets handle special casing in a followup pr

tlambert03 (approved):
good stuff.  i'm inclined to merge as is, and then do some refactoring later today.  sound good?

fdrgsp:
> good stuff. i'm inclined to merge as is, and then do some refactoring later today. sound good?

sounds great!!!

PR #222, opened 2022-12-04
fix: make sure toolbar is always on top

tlambert03:
this PR adds an eventFilter to the toolbar widget that redocks it at the top when it detects that it has been inserted into the main window (which will currently always be on the right side for napari plugins)


tlambert03:
merging, but cc @fdrgsp  ... just wanted you to see/learn about `eventFilter`s

PR #224, opened 2022-12-04
Remove seq meta

tlambert03:
this PR removes the `SEQUENCE_META` dict altogether, in favor of putting the metadata in the `MDASequence.metadata` dict which we now have thanks to @ianhi  in https://github.com/pymmcore-plus/useq-schema/pull/35


fdrgsp (approved):
Very nice!

PR #225, opened 2022-12-04
feat: add scale to layers

fdrgsp:
This PR adds a `layer.scale` info to all the image layers acquire through `napari-micromanager` (mda, explorer, preview).

### DIFF
--- a/src/napari_micromanager/main_window.py
+++ b/src/napari_micromanager/main_window.py
@@ -303,7 +303,6 @@ class MainWindow(MicroManagerToolbar):
             layer.scale = self._get_scale_from_sequence(
                 sequence, layer.data.shape, meta
             )
-            print("#############", layer.scale, layer.data.shape)
 
             # add metadata to layer
             layer.metadata["mode"] = meta.mode
@@ -368,7 +367,7 @@ class MainWindow(MicroManagerToolbar):
         # sourcery skip: use-contextlib-suppress
         try:
             index = sequence.used_axes.index("z")
-            if meta.split_channels and sequence.axis_order == "tpcz":
+            if meta.split_channels and sequence.axis_order in ["tpcz", "ptcz"]:
                 index -= 1
             scale[index] = getattr(sequence.z_plan, "step", 1)
         except ValueError:
--- a/src/napari_micromanager/main_window.py
+++ b/src/napari_micromanager/main_window.py
@@ -4,7 +4,7 @@ import atexit
 import contextlib
 import tempfile
 from collections import defaultdict
-from typing import TYPE_CHECKING, Any, Tuple
+from typing import TYPE_CHECKING, Any, List
 
 import napari
 import numpy as np
@@ -356,7 +356,7 @@ class MainWindow(MicroManagerToolbar):
 
     def _get_scale_from_sequence(
         self, sequence: MDASequence, layer: napari.layers.Image
-    ) -> Tuple[float, ...]:  # noqa: U006
+    ) -> List[float]:  # noqa: U006
         """Calculate and return the layer scale.
 
         ...using pixel size, layer shape and the MDASequence z info.
@@ -369,7 +369,7 @@ class MainWindow(MicroManagerToolbar):
                 scale[-3] = sequence.z_plan.step
             elif sequence.axis_order.index("z") == 2:
                 scale[-4] = sequence.z_plan.step
-        return tuple(scale)
+        return scale
 
     @ensure_main_thread  # type: ignore [misc]
     def _on_mda_frame(self, image: np.ndarray, event: ActiveMDAEvent) -> None:
--- a/src/napari_micromanager/main_window.py
+++ b/src/napari_micromanager/main_window.py
@@ -355,14 +355,10 @@ class MainWindow(MicroManagerToolbar):
         """
         scale = [1.0] * len(layer_shape)
         scale[-2:] = [self._mmc.getPixelSizeUm()] * 2
-        # sourcery skip: use-contextlib-suppress
-        try:
-            index = sequence.used_axes.index("z")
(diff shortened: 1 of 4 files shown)


src/napari_micromanager/main_window.py
```diff
@@ -344,6 +354,23 @@ def _get_defaultdict_layers(self, event: ActiveMDAEvent) -> defaultdict[Any, set
                 layergroups[key].add(lay)
         return layergroups
 
+    def _get_scale_from_sequence(
+        self, sequence: MDASequence, layer: napari.layers.Image
+    ) -> Tuple[float, ...]:  # noqa: U006
```
tlambert03, line 359: i _think_ you should only need to use U006 when you're doing a module level type annotation... but all function parameters and return annotations shouldn't need them?  what happens if you remove?  (believe it will just cast it to `tuple[]` which should be fine here, since we have `from __future__ ... `

src/napari_micromanager/main_window.py
```diff
@@ -344,6 +354,23 @@ def _get_defaultdict_layers(self, event: ActiveMDAEvent) -> defaultdict[Any, set
...
+        """
+        scale = [self._mmc.getPixelSizeUm()] * len(layer.data.shape)
+        if len(sequence.z_plan) > 1 and isinstance(
+            sequence.z_plan, (ZTopBottom, ZRangeAround, ZAboveBelow)
+        ):
+            if sequence.axis_order.index("z") == 3:
+                scale[-3] = sequence.z_plan.step
+            elif sequence.axis_order.index("z") == 2:
+                scale[-4] = sequence.z_plan.step
+        return tuple(scale)
```
tlambert03, line 372: couple thoughts:

- `scale = [self._mmc.getPixelSizeUm()] * len(layer.data.shape)` only makes sense if there will _only_ ever be 2 or 3 dimensions to `layer.data`, otherwise you're also scaling the time dimension, etc...  `use [1] * len()` and then only set the last two to the pixel size
- avoid using `list.index('something')` twice in the same block.  Just index it once and use it twice
- i think you want `sequence.used_axes` here instead of axis_order.
so, something like this?

```suggestion
        scale = [1] * len(layer.data.shape)
        scale[-2:] = [self._mmc.getPixelSizeUm()] * 2
        try:
            index = sequence.used_axes.index('z')
            scale[index] = getattr(sequence.z_plan, 'step', 1)
        except ValueError:
            pass # z not in used_axes afterall ?
        return tuple(scale)
```

src/napari_micromanager/main_window.py
```diff
@@ -329,12 +335,16 @@ def _add_explorer_positions_layers(
...
+            # set layer scale
+            layer.scale = self._get_scale_from_sequence(sequence, layer)
+
             # add metadata to layer
             layer.metadata["mode"] = meta.mode
             layer.metadata["useq_sequence"] = sequence
             layer.metadata["uid"] = sequence.uid
             layer.metadata["grid"] = pos.split("_")[-3]
             layer.metadata["grid_pos"] = pos.split("_")[-2]
+            layer.metadata["scale"] = layer.scale
```
tlambert03, line 347: is this "just in case" or do you use it elsewhere?  adding `layer.scale` to `layer.metadata['scale']` feels like an unnecessary duplication? (we already have it at layer.scale, no?)

src/napari_micromanager/main_window.py
fdrgsp: however we need to differentiate when we split the channels...otherwise the scale is not correct.

for example, if we have 2 channels and a z stack with step size 2µm, we get:
- scale = [1. 1. 2. 1.] with layer shape = (1, 3, 512, 512) 
when actually should be:
- scale = [1. 2. 1. 1.] with layer shape = (1, 3, 512, 512) 

that's why I used `sequence.axis_order.index("z")` and changing the index like below:
```
 ...
 index = sequence.axis_order.index("z")
        if index == 2:
                index = -4
        elif index == 3:
                index = -3
...
```
What do you think?

src/napari_micromanager/main_window.py
fdrgsp: no you're right, is a duplication.

src/napari_micromanager/main_window.py
fdrgsp: I'll write a proper test soon

tests/test_layer_scale.py
```diff
@@ -0,0 +1,70 @@
...
+
+if TYPE_CHECKING:
+    from pytestqt.qtbot import QtBot
+
+
+@pytest.mark.parametrize("Z", ["", "withZ"])
+@pytest.mark.parametrize("splitC", ["", "splitC"])
+@pytest.mark.parametrize("C", ["", "withC"])
+@pytest.mark.parametrize("T", ["", "withT"])
+def test_layer_scale(qtbot: QtBot, main_window: MainWindow, T, C, splitC, Z):
```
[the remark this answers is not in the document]
tlambert03: think about "unit tests" and look at the changed lines of the PR to guide you as to what should be tested.  In this case, you've added a new function that takes a sequence and a layer and returns a list.  You don't need to spin up an entire MDA to test that.  

In fact, this is probably better as a pure function that takes a sequence, a pixel size, and a shape tuple:
```python
def get_scale_from_sequence(
        sequence: MDASequence, shape: tuple[int, ...]
    ) -> list[float]:
```
and then either tack on the pixel size afterwards, or add it as a `float` parameter.

Then it's clear you don't need to run a full mda to test it.  just build a sequence and a shape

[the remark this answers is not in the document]
fdrgsp: The mda is not actually running, I'm only calling the `main_window._on_mda_started(mda)` method to create the `zarr` layer and then asserting that the scale is correct for each tpzc experiment. Is it not similar to what you suggested?

tlambert03 (approved):

PR #231, opened 2022-12-06
refactor: pull out mda handling logic into MDAHandler object

tlambert03:
trying to isolate the parts of the main window that just mediate events between the core and napari (but which aren't a widget)


tlambert03:
ok, this PR is ready. (cc @fdrgsp  just so you're aware)

the point of this PR is to
1. separate all the logic that handles receiving frames from an mda and updating the napari viewer from the MainWindow: since it doesn't really need to be associated with a widget, and it's becoming large.  This lets us test things without even needing to create a main window (note the change to the layer scale test)
2. deduplicate more logic that is shared between the mda and the explorer.

PR #235, opened 2022-12-07
refactor: simplify on-frame logic

tlambert03:
continueing in the simplification/dedup of the MDAHandler logic.  this unifies the `on_mda_frame` behavior.    It adds a new  stateless `_id_idx_layer` function that takes an event object and returns the info needed to get the appropriate zarr array, the index in the array at which the event image belongs, and teh name of the corresponding layer in the viewer.

### DIFF
--- a/src/napari_micromanager/_mda_handler.py
+++ b/src/napari_micromanager/_mda_handler.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import contextlib
 import tempfile
 from collections import defaultdict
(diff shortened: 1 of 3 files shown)


tlambert03:
tests are passing, but @fdrgsp, it would be good if you could pull this and run it locally, checking all the things you manually check with the explorer (such as translation, etc...)  If you find anything that has broken, let's discuss it and write a test to assert it

PR #242, opened 2023-01-18
Fix: Update mda for new pymmcore-widgets

ianhi:
Prior to this change the MDA widget would not launch due to a menagerie of naming errors.

This includes part of the changes from #237, but not the changes to the explorer tests.

attn: @JasonYu1 and @fdrgsp


tlambert03:
thank you @ianhi  ... i will merge this today, but want to get test running locally first... it _may_ require doing some stuff that supersedes this PR, but I'll make sure to merge it into whatever solution gets used

PR #246, opened 2023-01-19
fix: add fixes for new widgets

tlambert03:
pins widgets to >0.3.0 and gets tests passing again.

Note @fdrgsp  ... I'm just skipping the sample explorer tests here.  Please fix those when you get a chance


ianhi:
nice!

PR #257, opened 2023-02-24
feat: Allow passing a config from command line

### LINKED ISSUE #252: Allow taking config file as command line argument
i.e. `napari-micromanager my_config.cfg` should auto load the config file.

ianhi:
Allows for

```
napari-micromanager --config path/to/config.cfg
```

closes #252


src/napari_micromanager/main_window.py
```diff
@@ -62,6 +65,13 @@ def __init__(self, viewer: napari.viewer.Viewer) -> None:
         self.destroyed.connect(self._cleanup)
         atexit.register(self._cleanup)
 
+        if config is not None:
+            try:
+                self._mmc.loadSystemConfiguration(config)
+            except FileNotFoundError:
+                # don't crash if the user passed an invalid config
```
tlambert03, line 72: should warn though, or somehow indicate that a passed parameter was ignored

tlambert03:
happy to get this in @ianhi  ... but there are two comments that need addressing and merge conflicts to fix

ianhi:
> but there are two comments that need addressing and merge conflicts to fix

Yup, I'm catching up on some github things this week. hope to address this and several other things soon.

tlambert03:
oh actually, I can reproduce it if I run `python -m pytest -v --color=yes --cov=napari_micromanager --cov-report=xml` ... (like we do on CI).  I think you need to touch up the sys.argv parsing

tests/test_main.py
```diff
@@ -1,18 +1,31 @@
...
+    [[""], ["", str(Path(__file__).parent / "test_config.cfg")], ["", "nonexistant"]],
+)
+def test_cli_main(argv: list) -> None:
     import napari
     from napari.qt import QtViewer
 
     with patch("napari.run") as mock_run:
         with patch("qtpy.QtWidgets.QMainWindow.show") as mock_show:
-            main()
+            with patch("napari.utils.notifications.show_warning") as mock_warning:
```
ianhi, line 19: It seems that this patch isn't actually patching (i.e. the warning still gets displayed) and it is never called.

Any idea what I'm doing wrong here?

ianhi:
> oh actually, I can reproduce it if I run python -m pytest -v --color=yes --cov=napari_micromanager --cov-report=xml ... (like we do on CI). I think you need to touch up the sys.argv parsing (or modify the test)

This was very helpful thank you!

tests/test_main.py
tlambert03: not immediately, but you know what... just use `warnings.warn()`, napari will intercept that anyway and show a warning in the UI (or at least it used to?)

ianhi: hmm it doesn't seem to capture it in the same way. maybe i need to change the level

tlambert03: oh, it's possible that the napari warnings hook hasn't yet been invoked by this point?

ianhi: actually scratch that, now the show warnign is also not showing, so maybe just going away to quickly. it still shows up in teh terminal

tlambert03:
i finished this up

PR #258, opened 2023-02-24
feat: use another thread to write to mda zarr

ianhi:
Fixes: https://github.com/pymmcore-plus/napari-micromanager/issues/256

Untested because i'm not yet sure this is the fully correct way to do this. Would be great to get some eyes on this from @tlambert03 and @fdrgsp 

This tries to do as much of the work of processing an MDA frame in a third thread as possible. I also wanted to ensure that the most recently captured image is the most recently displayed, so this end `deque` for its LIFO behavior and because it is higher performance than queue (https://bugs.python.org/issue15329#msg199368). 

Then when the worker thread has a chance it processes all the previous indices that it didn't get a chance to deal with before a new image came in. This is the source of the complexity of checking `_largest_idx`

To user test:

1. Open napari-micromanager 
2. Load the demo config
3. Set up MDA with one channel - 100 time points at 0 interval
4. run - check that gui is functional

### DIFF
--- a/src/napari_micromanager/_mda_handler.py
+++ b/src/napari_micromanager/_mda_handler.py
@@ -195,7 +195,6 @@ class _NapariMDAHandler:
             layer: Image = self.viewer.layers[layer_name]
             if not layer.visible:
                 layer.visible = True
-            layer.reset_contrast_limits()
 
     def _on_mda_finished(self, sequence: MDASequence) -> None:
         # Save layer and add increment to save name.
--- a/src/napari_micromanager/_mda_handler.py
+++ b/src/napari_micromanager/_mda_handler.py
@@ -198,9 +198,9 @@ class _NapariMDAHandler:
         # Save layer and add increment to save name.
         self._mda_running = False
         if (meta := sequence.metadata.get(SEQUENCE_META_KEY)) is not None:
-            # TODO: make sure io thread finishes before saving
-            # thread workers don't seem to have .join method?
-            # self._io_t.join()
+            # this should be ok because we fully empty the deck before this
+            # saving. Also in the future below will be removed in favor of
+            # proper writer infrastructure.
             sequence = cast("ActiveMDASequence", sequence)
             save_sequence(sequence, self.viewer.layers, meta)
 
--- a/src/napari_micromanager/_mda_handler.py
+++ b/src/napari_micromanager/_mda_handler.py
@@ -2,15 +2,24 @@ from __future__ import annotations
 
 import contextlib
 import tempfile
+import time
 from collections import defaultdict, deque
-from typing import TYPE_CHECKING, Any, Callable, Iterable, Iterator, Sequence, cast
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Deque,
+    Generator,
+    Iterable,
+    Iterator,
+    Sequence,
+    cast,
+)
 
 import napari
 import zarr
 from napari.experimental import link_layers, unlink_layers
-from pymmcore_plus import CMMCorePlus
 from superqt.utils import create_worker, ensure_main_thread
-from useq import MDAEvent, MDASequence
 
 from ._mda_meta import SEQUENCE_META_KEY, SequenceMeta
 from ._saving import save_sequence
@@ -74,7 +83,7 @@ class _NapariMDAHandler:
 
         # mapping of id -> (zarr.Array, temporary directory) for each layer created
         self._tmp_arrays: dict[str, tuple[zarr.Array, tempfile.TemporaryDirectory]] = {}
-        self._deck: deque = deque()
+        self._deck: Deque[tuple[np.ndarray, MDAEvent]] = deque()  # noqa: UP006
 
         # Add all core connections to this list.  This makes it easy to disconnect
         # from core when this widget is closed.
@@ -139,20 +148,18 @@ class _NapariMDAHandler:
         self._mda_running = True
         # init index will always be less than any event index
         self._largest_idx: tuple[int, ...] = (-1,)
-        self._io_t = create_worker(self._watch_mda, _start_thread=True)
+        self._io_t = create_worker(
+            self._watch_mda,
+            _start_thread=True,
+        )
 
-    def _watch_mda(self) -> None:
+    def _watch_mda(self) -> Generator[None, None, None]:
         while self._mda_running:
-            try:
+            if self._deck:
                 self._process_frame(*self._deck.pop())
-            except IndexError:
-                pass
-        import time
-
-        time.sleep(0.5)
-        # Add any images earlier images we may have skipped for performance
-        while self._deck:
-            self._process_frame(*self._deck.pop())
+            else:
+                time.sleep(0.1)
+            yield
 
     def _on_mda_frame(self, image: np.ndarray, event: MDAEvent) -> None:
         """Add the newest frame to the deck."""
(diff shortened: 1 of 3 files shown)


src/napari_micromanager/_mda_handler.py
```diff
@@ -163,11 +194,15 @@ def _on_mda_frame(self, image: np.ndarray, event: MDAEvent) -> None:
...
-            # layer.reset_contrast_limits()
+            layer.reset_contrast_limits()
 
     def _on_mda_finished(self, sequence: MDASequence) -> None:
         # Save layer and add increment to save name.
+        self._mda_running = False
         if (meta := sequence.metadata.get(SEQUENCE_META_KEY)) is not None:
+            # TODO: make sure io thread finishes before saving
+            # thread workers don't seem to have .join method?
+            # self._io_t.join()
```
ianhi, line 205: @tlambert03 am I missing something here?

src/napari_micromanager/_mda_handler.py
```diff
@@ -132,11 +133,31 @@ def _on_mda_started(self, sequence: MDASequence) -> None:
...
+    def _watch_mda(self) -> None:
+        print("starting to watch")
+        while self._mda_running:
+            try:
+                self._process_frame(*self._deck.pop())
+            except IndexError:
+                pass
+        import time
+
+        time.sleep(0.5)
```
ianhi, line 151: todo: Make a real choice for this number. I chose this based on what was convenient for development. Ideally it would be also be adjustable - open to suggestions.

src/napari_micromanager/_mda_handler.py
```diff
@@ -132,11 +133,31 @@ def _on_mda_started(self, sequence: MDASequence) -> None:
...
-                return
 
-    @ensure_main_thread  # type: ignore [misc]
+        self._mda_running = True
+        # init index will always be less than any event index
+        self._largest_idx: tuple[int, ...] = (-1,)
+        self._io_t = create_worker(self._watch_mda, _start_thread=True)
+
+    def _watch_mda(self) -> None:
+        print("starting to watch")
```
ianhi, line 143:
```suggestion
```

woops

src/napari_micromanager/_mda_handler.py
```diff
@@ -132,11 +133,31 @@ def _on_mda_started(self, sequence: MDASequence) -> None:
...
+        while self._mda_running:
+            try:
+                self._process_frame(*self._deck.pop())
+            except IndexError:
+                pass
```
ianhi, line 148: maybe sleep for an extra long time here?

src/napari_micromanager/_mda_handler.py
```diff
@@ -163,11 +194,15 @@ def _on_mda_frame(self, image: np.ndarray, event: MDAEvent) -> None:
             layer: Image = self.viewer.layers[layer_name]
             if not layer.visible:
                 layer.visible = True
-            # layer.reset_contrast_limits()
+            layer.reset_contrast_limits()
```
ianhi, line 197: accidental uncomment
```suggestion
            # layer.reset_contrast_limits()
```

src/napari_micromanager/_mda_handler.py
```diff
@@ -150,11 +170,21 @@ def _on_mda_frame(self, image: np.ndarray, event: MDAEvent) -> None:
...
-        self.viewer.dims.current_step = tuple(cs)
 
+        if im_idx > self._largest_idx:
+            self._largest_idx = im_idx
+            # Processing the most recent event
+            # update the viewer in the main thread
+            cs = list(self.viewer.dims.current_step)
+            for a, v in enumerate(im_idx):
+                cs[a] = v
+            self._update_viewer_dims(cs, layer_name, event)
```
tlambert03, line 181: slightly wary of accessing `self.viewer` from a thread.  might want some locks in here.  Also, I wonder it would be possible for the thing triggering `_update_viewer_dims` to not be `_process_frame`

src/napari_micromanager/_mda_handler.py
tlambert03: no, `QRunnables` are slightly more restricted than true QThreads.  They don't have a `join()`... but they do have an `is_running()` method that you could wait on.  They also emit a `finished` signal that you could connect to a new method that does the saving.

```py
create_worker(self._watch_mda, _start_thread=True, _connect={'finished': self._on_io_finished})
```

tlambert03 (approved):
awesome!!  it's way better 🤩 

happy to go with this and use it for a while,

src/napari_micromanager/_mda_handler.py
ianhi: How about just moving that access in to the function that has `ensure_main_thread`?

tlambert03: sure, and then it would just be `self._update_viewer_dims(layer_name, event)`?  if that works too, I like that

src/napari_micromanager/_mda_handler.py
ianhi: Is there an advantage to QRunnables that comes with the restrictions? I'm not sure I fully grasp the breakdown of thread vs runnable

src/napari_micromanager/_mda_handler.py
```diff
@@ -72,6 +72,7 @@ def __init__(self, mmcore: CMMCorePlus, viewer: napari.viewer.Viewer) -> None:
 
         # mapping of id -> (zarr.Array, temporary directory) for each layer created
         self._tmp_arrays: dict[str, tuple[zarr.Array, tempfile.TemporaryDirectory]] = {}
+        self._deck: deque = deque()
```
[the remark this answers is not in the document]
ianhi: good call. Also for the future we should go to just collections.deque (at least once napari-micro is python 3.9+ only)

https://docs.python.org/3/library/typing.html#typing.Deque

> Deprecated since version 3.9: [collections.deque](https://docs.python.org/3/library/collections.html#collections.deque) now supports subscripting ([]). See [PEP 585](https://peps.python.org/pep-0585/) and [Generic Alias Type](https://docs.python.org/3/library/stdtypes.html#types-genericalias).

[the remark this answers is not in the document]
tlambert03: Yep, all the generic collections can be updated then.  3.8 end of life is Oct 2024

ianhi:
working on rebasing. but running into what seem like some issues in pymmcore-widgets:

```
    def _enable_run_btn(self) -> None:
        """Enable run button.
    
        ...if there is a channel group and a preset selected or the channel checkbox
        is checked and there is at least one channel selected.
        """
        if self._mmc.getChannelGroup() and self._mmc.getCurrentConfig(
            self._mmc.getChannelGroup()
        ):
            if self.ch_cbox.isChecked() and not self.channel_widget._table.rowCount():
                self.buttons_wdg.run_button.setEnabled(False)
            else:
                self.buttons_wdg.run_button.setEnabled(True)
>       elif not self.ch_cbox.isChecked() or not self.channel_widget._table.rowCount():
E       AttributeError: 'QWidget' object has no attribute 'isChecked'
```

attn @fdrgsp

ianhi:
@tlambert03 I believe I addressed all review comments - this should be ok.

I went ahead with this because it was pretty close and is more self contained than  #264 which has several unrelated changes making it trickier to parse through. If this one does get merged then I would very much support a follow up PR with performance enhancements (as isolated changes ideally).

tlambert03:
trying to get tests to run

tests/test_layer_scale.py
fdrgsp, line 1: I would make sure we are not saving the files in this test, I would switch from
```py
sequence = mda_sequence_splits.replace(axis_order=axis_order)
```
to
```py
    sequence = mda_sequence_splits.replace(
        axis_order=axis_order,
        metadata={SEQUENCE_META_KEY: SequenceMeta(should_save=False)},
    )
```

fdrgsp:
>  #264 which has several unrelated changes making it trickier to parse through.

@ianhi yes that's true sorry about that:). Maybe you can ignore everything else and just have a look at the `tests` and the `_mda_handler.py` which are the only things that are important for this PR. I did test it on the microscope and works quite well without freezing the gui.

one think I found useful and I would add is to specifying the chunk size when creating the zarr since it improves the image assignment when they are popped from the _deck (https://github.com/fdrgsp/napari-micromanager/blob/3256672406344d387f8151fcfbc4b1214e96fd19/src/napari_micromanager/_mda_handler.py#L128)

I also added a comment about tests here: 

Anyway I think this PR is ok, I did not tested it on a real microscope yet but It's close to #264 which I did test.

tlambert03:
does #272 close this?

ianhi:
superseded by #272

PR #259, opened 2023-03-29
fix: don't leave scale undefined

ianhi:
Follow up to #244 in which I fixed one bug, but introduced another


ianhi:
rebased on https://github.com/pymmcore-plus/napari-micromanager/pull/260 to hopefully fix the tests

PR #264, opened 2023-07-21
feat: Update thread to write to mda zarr

fdrgsp:
followup of #258.

In this PR I've updated the code with the suggestions in #258 + I've added a method (`_process_remaining_frames`) that is used to add to the zarr layer(s) any frames remaining in  `_deck` after the acquisition is finished. Note that at the moment, since we don't have a proper writer in `napari-micromanager` this method is called by `_on_mda_finished` but once we have a writer, we can simply link this method to `_io_thread` `_connect` (`"finished": self._process_remaining_frames`).

To speed up the process it was also necessary to specify the chunk size when creating the zarr.

I did few test on our microscope and this way of handling the acquired images works great!

In addition I've added some fixes for the use of the current `pymmcore-widgets` (main git branch).
Test failing until new release of `pymmcore-widgets`.

closes #258.


tlambert03:
The private _sequence module definitely shouldn't be imported here.  Let's talk about what you need to accomplish and make it public if there's no other way

fdrgsp:
> The private _sequence module definitely shouldn't be imported here. Let's talk about what you need to accomplish and make it public if there's no other way

Sure :) I'm doing some tests with your sequencable PR and it's awesome!!! I pushed this commit to not loose the testing progress :)

ianhi:
> In addition I've added some fixes for the use of the current pymmcore-widgets (main git branch).

@fdrgsp can you please make this a separate self contained PR. Wrapping it up in a big PR like this makes it difficult for me to work on other things (and led to me thinking this was an unfixed bug: https://github.com/pymmcore-plus/napari-micromanager/issues/268)

fdrgsp:
closing. Open a simpler version #272.

PR #266, opened 2023-07-26
fix: remove defunct sample explorer

ianhi:
SampleExplorer was removed in: https://github.com/pymmcore-plus/pymmcore-widgets/pull/139

need to also remove it here. Also seems as though some recent changes in useq have led to mypy issues


ianhi:
huh. locally I got these mypy errors, but not here it seems:

```
src/napari_micromanager/_saving.py:86: error: Argument 1 to "len" has
incompatible type
"Union[MultiPhaseTimePlan, TIntervalDuration, TIntervalLoops, TDurationLoops, None]";
expected "Sized"  [arg-type]
                    if len(sequence.time_plan) == 0
                           ^~~~~~~~~~~~~~~~~~
src/napari_micromanager/_saving.py:109: error: Argument 1 to "len" has
incompatible type
"Union[MultiPhaseTimePlan, TIntervalDuration, TIntervalLoops, TDurationLoops, None]";
expected "Sized"  [arg-type]
    ...      ax = sequence.axis_order.index("p") if len(sequence.time_plan) >...
                                                        ^~~~~~~~~~~~~~~~~~
Found 2 errors in 1 file (checked 1 source file)
```

ianhi:
I think merging this will allow https://github.com/pymmcore-plus/napari-micromanager/pull/258 to be merged also

tlambert03 (approved):
thanks @ianhi!

tlambert03:
> huh. locally I got these mypy errors, but not here it seems:

yep, these are valid... looks like pre-commit is perhaps using a cached environment.  Will fix elsewhere

PR #269, opened 2023-07-27
fix: use widget from pymmcore-widgets main branch

fdrgsp:
@ianhi linked to https://github.com/pymmcore-plus/napari-micromanager/pull/264#issuecomment-1653752822
This PR updates `napari-micromanager` using the `pymmcore-widgets` in main branch.


ianhi:
This raises the question of the dependency structure of the testing between these two repos. Would it maybe make more sense to do the following?

1. In regular CI use the main branch of pymmcore-widgets/pymmcore-plus
2. In testing prior to a release use the currently released versions

It seems that that would allow smooth development, but still make sure released versions are compatible.
@tlambert03 @fdrgsp

tlambert03:
the structure I prefer is for upstream libraries to test against the main branch of their dependents with `-W ignore`
- useq tests main branch of pymmcore-plus
- pymmcore-plus tests main branch of widgets
- widgets tests main branch of napari-micromanager

While the downstream libraries test against the current release (NOT against the main branch).
Once the tests are in place, it's the upstream libraries responsibility to emit warnings rather than breakages, and let the downstream libraries test fail with a warning (rather than an actual error).  This mimics the case that would happen with a downstream third party library we're not testing.

I know you've just jumped into an odd state, but we're actively working on this and you just caught us at a funny time.

src/napari_micromanager/_gui_objects/_mda_widget.py
```diff
@@ -22,36 +30,35 @@ def __init__(
...
     def _toggle_split_channel(self) -> None:
-        if not self.channel_groupbox.value():
-            self.checkBox_split_channels.setChecked(False)
-
-    def _toggle_checkbox_save_pos(self) -> None:
         if (
-            self.position_groupbox.isChecked()
-            and len(self.position_groupbox.value()) > 0
+            not self.channel_widget.value()
+            or self.channel_widget._table.rowCount() == 1
```
tlambert03, line 56: this is a usage of a private widget in pymmcore-widgets.  If there is literally no public way to access this and you need it here, then you should make it public in pymmcore-widgets

tlambert03:
Hooray, seeing some green lights here

PR #272, opened 2023-08-04
feat: Update thread to write to mda zarr

fdrgsp:
@ianhi  I closed the old PR (#264) and open this one that only contains the changes relevant to the thread so that it is easier for you to have a look.

followup of https://github.com/pymmcore-plus/napari-micromanager/pull/258.

In this PR I've updated the code with the suggestions in https://github.com/pymmcore-plus/napari-micromanager/pull/258 + I've added a method (_process_remaining_frames) that is used to add to the zarr layer(s) any frames remaining in  _deck after the acquisition is finished. Note that at the moment, since we don't have a proper writer in napari-micromanager this method is called by _on_mda_finished but once we have a writer, we can simply link this method to _io_thread _connect ("finished": self._process_remaining_frames).

To speed up the process it was also necessary to specify the chunk size when creating the zarr.

I did few test on our microscope and this way of handling the acquired images works great!

### DIFF
--- a/src/napari_micromanager/_mda_handler.py
+++ b/src/napari_micromanager/_mda_handler.py
@@ -399,7 +399,7 @@ def _id_idx_layer(event: ActiveMDAEvent) -> tuple[str, tuple[int, ...], str]:
         try:
             im_idx += (event.index[k],)
         # if axis not in event.index
-        # e.g. if we have bot a position sequence grid and a single position
+        # e.g. if we have both a position sequence grid and a single position
         except KeyError:
             im_idx += (0,)
 
--- a/src/napari_micromanager/_mda_handler.py
+++ b/src/napari_micromanager/_mda_handler.py
@@ -289,12 +289,12 @@ class _NapariMDAHandler:
         )
 
 
-def _get_axis_labels(sequence: MDASequence) -> tuple[list[str], bool]:
+def _get_axis_labels(sequence: MDASequence) -> list[str]:
     """Get the axis labels using only axes that are present in events."""
     # axis main sequence
     main_seq_axis = list(sequence.used_axes)
     if not sequence.stage_positions:
-        return main_seq_axis, False
+        return main_seq_axis
     # axes from sub sequences
     sub_seq_axis: list = []
     for p in sequence.stage_positions:
@@ -302,7 +302,12 @@ def _get_axis_labels(sequence: MDASequence) -> tuple[list[str], bool]:
             sub_seq_axis.extend(
                 [ax for ax in p.sequence.used_axes if ax not in main_seq_axis]
             )
-    return main_seq_axis + sub_seq_axis, bool(sub_seq_axis)
+    return main_seq_axis + sub_seq_axis
+
+
+def _has_sub_sequences(sequence: MDASequence) -> bool:
+    """Return True if any stage positions have a sub sequence."""
+    return any(p.sequence is not None for p in sequence.stage_positions)
 
 
 def _determine_sequence_layers(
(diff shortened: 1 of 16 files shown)


src/napari_micromanager/_mda_handler.py
```diff
@@ -107,15 +111,20 @@ def _on_mda_started(self, sequence: MDASequence) -> None:
...
             tmp = tempfile.TemporaryDirectory()
             dtype = f"uint{self._mmc.getImageBitDepth()}"
-
             # create the zarr array and add it to the viewer
-            z = zarr.open(str(tmp.name), shape=shape + yx_shape, dtype=dtype)
+            z = zarr.open(
+                str(tmp.name),
+                shape=shape + yx_shape,
+                dtype=dtype,
+                chunks=tuple([1] * len(shape) + yx_shape),  # VERY IMPORTANT FOR SPEED!
```
ianhi, line 126: Would it be more optimal to have the shape be something like the length of a z-stack? 

also inthe future with the grouped acq events may be worth revisiting, but I htink this is a good start

src/napari_micromanager/_mda_handler.py
```diff
@@ -290,20 +396,29 @@ def _id_idx_layer(event: ActiveMDAEvent) -> tuple[str, tuple[int, ...], str]:
...
     _id = f"{event.sequence.uid}{suffix}"
 
     # the index of this event in the full zarr array
-    im_idx = tuple(event.index[k] for k in axis_order)
+    im_idx: tuple[int, ...] = ()
+    for k in axis_order:
+        try:
+            im_idx += (event.index[k],)
+        # if axis not in event.index
+        # e.g. if we have bot a position sequence grid and a single position
```
ianhi, line 417:
```suggestion
        # e.g. if we have both a position sequence grid and a single position
```

im not sure I fully understand this comment. Can you give an example of how to create a sequence that would trigger this?

src/napari_micromanager/_mda_handler.py
```diff
@@ -213,14 +289,36 @@ def _create_empty_image_layer(
         )
 
 
+def _get_axis_labels(sequence: MDASequence) -> list[str]:
```
ianhi, line 292: This should probably live in `_util`?

ianhi (approved):
thanks for picking this up and bringing it to the finish line!

src/napari_micromanager/_mda_handler.py
fdrgsp: @ianhi yes maybe! I did some test trying to display a `sequenced` acquisition in real-time and setting this to the shape of a single image gave me the fastest result. But we can definitely do more tests!

src/napari_micromanager/_mda_handler.py
fdrgsp: this is an example of a `MDASequence` that I found it needs something like this try/except:

```py
seq = MDASequence(
    channels=["DAPI"],
    stage_positions=[
        {"x": 0, "y": 0, "z": 0, "sequence": {"grid_plan": {"rows": 2, "columns": 2}}},
        {"x": 0, "y": 0, "z": 0}
    ],
)
```
In this case, the `g` index is present when iterating through the _first position_ but **not** in the _second position_ because it does not have a `grid_plan` sub-sequence.

src/napari_micromanager/_mda_handler.py
fdrgsp: Yes we can move it there I think!

tlambert03:
hi folks.  good to see some activity here.  so, what's the status?  are we waiting on action in response to the comments above? or is this good to go?

ianhi:
> hi folks. good to see some activity here. so, what's the status? are we waiting on action in response to the comments above? or is this good to go?

good to go! remaining comment is mostly a musing and a maybe future improvement

tlambert03 (approved):
thanks!

PR #281, opened 2023-08-27
refactor: changing toolbar structure

### LINKED ISSUE #255: toolbar and mda window loads too wide
something about the size policies on the toolbar and the mda widget have them behaving wonky.  When you add the plugin, it makes the window wider than it needs to be, and the mda widget also is too large and needs to be resized each time.  Also its min-width might want to be smaller (so that resizing it doesn't feel like hitting a small target of goodness)

https://user-images.githubusercontent.com/1609449/220791579-d246e299-5dc8-4377-9481-2dc70b82c358.mov

tlambert03:
This refactors `_toolbar` to get rid of a lot of boilerplate and make it clearer what's going on.

However, some styles have been lost in the process (though I think those styles were a bit too "manually-curated" and specific).  So this will need a little work together with @fdrgsp to get back to a nice looking place.  I think we should do away with the group boxes, which are presumably only being used for the border (which could be added as a style to the toolbar, etc...)

fixes #255

### DIFF
--- a/src/napari_micromanager/_gui_objects/_toolbar.py
+++ b/src/napari_micromanager/_gui_objects/_toolbar.py
@@ -340,7 +340,6 @@ class PluginsToolBar(MMToolBar):
 
     def __init__(self, parent: MicroManagerToolbar) -> None:
         super().__init__("Plugins", parent)
-        self.addSubWidget(MMShuttersWidget())
         self.groupbox.setStyleSheet("border: 0px;")
 
         if not isinstance(parent, MicroManagerToolbar):
--- a/src/napari_micromanager/_gui_objects/_toolbar.py
+++ b/src/napari_micromanager/_gui_objects/_toolbar.py
@@ -21,7 +21,7 @@ from pymmcore_widgets import (
 from qtpy.QtCore import QEvent, QObject, QSize, Qt
 from qtpy.QtWidgets import (
     QDockWidget,
-    QGroupBox,
+    QFrame,
     QHBoxLayout,
     QLabel,
     QMainWindow,
@@ -42,9 +42,7 @@ from ._stages_widget import MMStagesWidget
 if TYPE_CHECKING:
     import napari.viewer
 
-TOOLBAR_SIZE = 45
 TOOL_SIZE = 35
-GROUPBOX_STYLE = "QGroupBox { border-radius: 3px; }"
 
 
 # Dict for QObject and its QPushButton icon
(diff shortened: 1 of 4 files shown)


tlambert03:
also, it's worth pointing out that the name `MicroManagerToolbar` is a bit misleading... since the widget itself is not a QToolBar, but rather it's a funny widget that simply waits to be added to a mainwindow, and then adds a bunch of toolbars to _that_ main window...  So the whole pattern probably could use a review

src/napari_micromanager/_gui_objects/_toolbar.py
```diff
@@ -394,3 +191,151 @@ def _add_dock_widget(
...
+
+    QPushButtons are connected to the `_show_dock_widget` method.
+
+    The QPushButton.whatsThis() property is used to store the key that
+    will be used by the `_show_dock_widget` method.
+    """
+
+    def __init__(self, parent: MicroManagerToolbar) -> None:
+        super().__init__("Plugins", parent)
+        self.addSubWidget(MMShuttersWidget())
```
fdrgsp, line 331: we added the MMShuttersWidget() in the ShuttersToolBar()
```suggestion

```

src/napari_micromanager/_gui_objects/_toolbar.py
```diff
@@ -394,3 +191,151 @@ def _add_dock_widget(
...
+        self.groupbox.setStyleSheet("border: 0px;")
+
+        if not isinstance(parent, MicroManagerToolbar):
+            raise TypeError("parent must be a MicroManagerToolbar instance.")
+
+        btn = QPushButton("MDA")
+        btn.setToolTip("MultiDimensional Acquisition")
+        btn.setMinimumHeight(TOOL_SIZE)
+        btn.setWhatsThis("MDA")
+        btn.clicked.connect(parent._show_dock_widget)
```
fdrgsp, line 341: adding the button to the toolbar
```suggestion
      btn.clicked.connect(parent._show_dock_widget)
      self.addSubWidget(btn)
```

src/napari_micromanager/_gui_objects/_toolbar.py
```diff
@@ -394,3 +191,151 @@ def _add_dock_widget(
...
         return dock_wdg
+
+
+# -------------- Toolbars --------------------
+
+
+class MMToolBar(QToolBar):
+    def __init__(
+        self, title: str, parent: QWidget = None, gb: QGroupBox | None = None
+    ) -> None:
```
fdrgsp, line 202: the `ConfigurationWidget()` was the last that had a QGroupBox. I removed it in https://github.com/pymmcore-plus/pymmcore-widgets/pull/185. after merging we can change this to:
```suggestion
class MMToolBar(QToolBar):
     def __init__(self, title: str, parent: QWidget = None) -> None:
```

src/napari_micromanager/_gui_objects/_toolbar.py
```diff
@@ -394,3 +191,151 @@ def _add_dock_widget(
...
+        self.groupbox.setStyleSheet(GROUPBOX_STYLE)
+        self.addWidget(self.groupbox)
+
+    def addSubWidget(self, wdg: QWidget) -> None:
+        self.groupbox.layout().addWidget(wdg)
+
+
+class ConfigToolBar(MMToolBar):
+    def __init__(self, parent: QWidget) -> None:
+        super().__init__("Configuration", parent, ConfigurationWidget())
```
fdrgsp, line 226: see `ConfigurationWidget()  `comment above
```suggestion
        super().__init__("Configuration", parent=parent)
        self._wdg = ConfigurationWidget()
        self.addSubWidget(self._wdg)
```
        self.addSubWidget(self._wdg)

PR #292, opened 2023-10-05
feat: use new pymmcore-widgets useq-widget

fdrgsp:
Use the new core-connected `pymmcore-widgets` `useq-widgets`.


tlambert03:
LGTM!

PR #309, opened 2024-02-01
[WIP] fix: fix lag in live mode after mda acquisition

fdrgsp:
I've noticed that after opening `napari-micromanager` the `live mode` works well without any lag. However, after running a MDA acquisition, the `live mode` starts to lag. See video below to see this behavior.

As a proposed solution, I modified the `CoreViewerLink` class and use the same `QTimer` pattern that we use in the `pymmcore-widgets` `ImagePreview `widget (https://github.com/pymmcore-plus/pymmcore-widgets/blob/d5d38c9cc24bc2ac99b5921ac2593a4037d258fb/src/pymmcore_widgets/_image_widget.py#L18).

---
CURRENT

https://github.com/pymmcore-plus/napari-micromanager/assets/70725613/3440ba20-afb3-46e9-8c94-8ab5a4863029

FIX

https://github.com/pymmcore-plus/napari-micromanager/assets/70725613/b4c790a2-697f-4103-9685-fca549364e65


src/napari_micromanager/_core_link.py
```diff
@@ -32,42 +35,46 @@ def __init__(
         self._mda_handler = _NapariMDAHandler(self._mmc, viewer)
         self._live_timer_id: int | None = None
 
+        self.streaming_timer = QTimer(parent=self)
```
tlambert03, line 38: I'm curious if you determined that using an independent `QTimer` instead of `self.startTimer` was sufficient to resolve the slow down?  I ask because this is a bit surprising to me.  Before accepting this change, I'd prefer to know *why* it matters.  Looking into the Qt docs for hints if need be.  If true, it might have to do with threading or something... but lets figure out why we're doing it

tlambert03:
I think that ultimately, the problem here is the `getExposure` with a Hamamatsu camera is returning an artificially high exposure time as of December 14th 2023.  See https://github.com/micro-manager/mmCoreAndDevices/issues/442

So I think this can be closed.

PR #310, opened 2024-02-02
fix: fix error when loading SequenceMeta as dict

fdrgsp:
fix `SequenceMeta` /`dict` error in tests not passing both locally and in in #309: `FAILED tests/test_multid_widget.py::test_saving_mda[splitC-nT=0-nZ=0-nC=1] - AttributeError: 'dict' object has no attribute 'split_channels'
`

note: new test added

### DIFF
--- a/src/napari_micromanager/_gui_objects/_mda_widget.py
+++ b/src/napari_micromanager/_gui_objects/_mda_widget.py
@@ -49,7 +49,7 @@ class MultiDWidget(MDAWidget):
             split_channels=bool(split),
             save_dir=widget_meta.get("save_dir", ""),
             file_name=widget_meta.get("save_name", ""),
-            should_save="save_dir" in widget_meta,
+            should_save=bool("save_dir" in widget_meta),
         )
         return sequence  # type: ignore[no-any-return]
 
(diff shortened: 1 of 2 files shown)


src/napari_micromanager/_gui_objects/_mda_widget.py
```diff
@@ -49,21 +49,20 @@ def value(self) -> MDASequence:
             split_channels=bool(split),
             save_dir=widget_meta.get("save_dir", ""),
             file_name=widget_meta.get("save_name", ""),
-            should_save=bool("save_dir" in widget_meta),
+            should_save="save_dir" in widget_meta,
```
tlambert03, line 52: did the linter remove this?  or did you?

tlambert03:
#311 was a more granular fix to the test.   the underlying issue has to do with useq schema's `replace()` method converting objects to dicts.  while this approach would technically work, it depends on the sequence first going through a widget `setValue` method.  So, it's not "wrong" per se, but it's a bit orthogonal to the core problem.  I'm going to reduce this down to a simple cast-dict-to-sequence and remove the rest of the PR

PR #314, opened 2024-02-05
fix: prevent preview update when mda is running

fdrgsp:
Necessary after https://github.com/pymmcore-plus/pymmcore-plus/pull/314

- use the new core.events.imageSnapped
- when a MDA is running, we don't want to update the image preview layer (since we're already updating the active experiment layers)
- add test

### DIFF
--- a/src/napari_micromanager/_core_link.py
+++ b/src/napari_micromanager/_core_link.py
@@ -35,7 +35,7 @@ class CoreViewerLink(QObject):
         # Add all core connections to this list.  This makes it easy to disconnect
         # from core when this widget is closed.
         self._connections: list[tuple[PSignalInstance, Callable]] = [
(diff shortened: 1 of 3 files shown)


src/napari_micromanager/_core_link.py
```diff
@@ -54,6 +54,12 @@ def cleanup(self) -> None:
     def timerEvent(self, a0: QTimerEvent | None) -> None:
         self._update_viewer()
 
+    def _image_snapped(self) -> None:
+        # If we are in the middle of an MDA, don't update the preview viewer.
+        if self._mda_handler._mda_running:
+            return
+        self._update_viewer(self._mmc.getImage())
```
tlambert03, line 61:
```suggestion
        if not self._mda_handler._mda_running:
            self._update_viewer(self._mmc.getImage())
```

PR #315, opened 2024-02-18
feat: remove save logic and use pymmcore-plus writer from MDAWidget

fdrgsp:
This PR only removes the saving logic that we had here in `napari-micromanager` and uses the `pymmcore_plus` writers implemented in the `MDAWidget` (https://github.com/pymmcore-plus/pymmcore-widgets/pull/279).

see also #316


src/napari_micromanager/_mda_handler.py
```diff
@@ -8,10 +8,11 @@
 
 import napari
 import zarr
+from pymmcore_plus.mda.handlers._util import get_full_sequence_axes
```
tlambert03, line 11: never import private stuff from other packages,  even if it's your package.  I know it's annoying.  but it just always comes back to bite you.  I'll update this one

tlambert03:
ok, i've had a look over and it looks good!

I did add back the  `test_saving_mda` test (we need to have at least one test to ensure we're saving things properly, even if it's not handled here).  But in doing so, I hit one error in the expected data shape.  You should see the error if you run:

`pytest -k 'test_saving_mda and no_splitC'`

tlambert03:
figured it out.  the mda_widget appears to be changing the `axis_order` of our requested mda.  looking into it

tlambert03:
fixes by https://github.com/pymmcore-plus/pymmcore-widgets/pull/286

tlambert03:
looks like we're now hitting this bug: https://github.com/pymmcore-plus/pymmcore-plus/issues/334

so there really is something to that one.  Will see if I can recreate in pymmcore-plus and fix

tlambert03:
bug is fixed upstream, all passing now, will merge

PR #325, opened 2024-03-11
docs: update image in docs

fdrgsp:
Update the `napari-micromanager` image in the docs to the recent version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 GUI interface between napari and micromanager powered by [pymmcore-plus](https://pymmcore-plus.github.io/pymmcore-plus/) and [pymmcore-widgets](https://pymmcore-plus.github.io/pymmcore-widgets/)
 
 ----------------------------------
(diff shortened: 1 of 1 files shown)


tlambert03:
great, yeah we need to update that.  However, can you use the same pattern as we had before?  (with the github url). 
The reason is that, while your PR will work fine for rendering the docs, the image would no longer work in the PyPI readme here: https://pypi.org/project/napari-micromanager/

tlambert03:
oh, and another thing that I've learned the hard way ... you want to be *very* careful about adding large assets like images to a github repo.  Once added, they become a _permanent_ part of your github repo history, and are very hard to remove from the history even if you later delete the file from HEAD.  The problem is that your git repo size is now bigger and cloning and all that stuff is slower... this file is huge at 7.62 MB, and would immediately quadruple the size of the repo.  So at the very least we'd want to compress the image a bit.  But using a URL to an image not inside the repo also fixes this problem ... so let's do that.  I'll open a new PR to show

PR #329, opened 2024-03-14
fix: Prevent lots of 'circular buffer empty' errors in MM

rixius:
Trying to read from an empty circular buffer causes lots of error messages in MM error log. Prevent them.

### DIFF
--- a/src/napari_micromanager/_core_link.py
+++ b/src/napari_micromanager/_core_link.py
@@ -77,6 +77,8 @@ class CoreViewerLink(QObject):
     def _update_viewer(self, data: np.ndarray | None = None) -> None:
         """Update viewer with the latest image from the circular buffer."""
         if data is None:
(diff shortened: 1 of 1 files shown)


tlambert03:
this was fine i think! why close?  cause of the test fails?

rixius:
Semantic check of the commit message failed, wanted to write it again with new commit message. Not necessary?
I have no idea about the other problem on macos.

tlambert03:
yeah, not necessary :)  just a silly little thing that I impose here,  by changing the title of the PR (as I did above) it will be fine.

tlambert03:
> I have no idea about the other problem on macos.

i think these are unrelated, and likely just due to some Qt issues when running on github actions that occasionally pop up.

tlambert03:
I just reopened this, so we should be good here.  thanks!!

PR #336, opened 2024-06-29
ci: fixing tests

tlambert03:


tlambert03:
wow... amazingly, just importing pymmcore before importing any Qt backend ended up fixing stuff on here. 😱

opened an issue at pymmcore: https://github.com/micro-manager/pymmcore/issues/119

but tests are passing here again as long as we ensure that we import pymmcore first

PR #338, opened 2024-07-16
test: fix tests, stop using make_napari_viewer fixture

fdrgsp:
add `pytest.mark.enable_console`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,12 +5,12 @@ ci:
 
 repos:
   - repo: https://github.com/crate-ci/typos
-    rev: v1.22.0
+    rev: v1.22.9
     hooks:
       - id: typos
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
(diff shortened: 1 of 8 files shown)


tlambert03:
I see this one remaining test fail a lot too.  Usually if I rerun it it goes away (will try now), but any idea what's going on there?

tlambert03:
Just curious, were you able to figure out where/when this marker got added in napari?

fdrgsp:
> Just curious, were you able to figure out where/when this marker got added in napari?

Nope, not sure what changed in napari's last week release that affected this behavior (for sure not the `make_napari_viewer` method though, it has not been recently modified)...

tlambert03:
for my tastes, this fix is to specific to something that napari is internally doing in their tests, and may well break again in the future.  I just think we shouldn't be using `make_napari_viewer`... will update

tlambert03:
pushed a change that stops using make_napari_viewer

PR #344, opened 2024-08-23
fix: shutter re-opening after changing config during live mode

simonecoppola:
fixes issue with shutter not re-opening


tlambert03 (approved):
Thanks @simonecoppola

PR #365, opened 2026-02-20
feat: swapping cores after launching the plugin & UniMMCore support

hinderling:
Is there any reason why not to use UniMMCore as default? Haven't found a good way to change the core after loading. Usecase is to load .cfg files with py devices (https://github.com/pymmcore-plus/pymmcore-plus/pull/531).


tlambert03:
> Is there any reason why not to use UniMMCore as default? 

just out of caution... not 100% sure how many people are using this plugin, and that was still considered experimental.  But, of course, its *designed* to be a drop-in replacement that falls back to all the same behavior, so in theory it should be just fine.

> Haven't found a good way to change the core after loading.

I actually think this would be the even better improvement to make here (and we can discuss making unicore default as well).  But it would really demonstrate good "core instance" hygiene here if we could swap the core out after launching the plugin.  Would you be willing to look into that?  basically:

- this is probably the first core creation moment:

    https://github.com/pymmcore-plus/napari-micromanager/blob/603e1ddf04f6784d020264375f3625d3c98eb575/src/napari_micromanager/_gui_objects/_toolbar.py#L66-L72

     but then it happens in the subclass again here:

    https://github.com/pymmcore-plus/napari-micromanager/blob/603e1ddf04f6784d020264375f3625d3c98eb575/src/napari_micromanager/main_window.py#L28-L37
 
    I think it would be great if there was one (and only one) method that creates/replaces the core.  It could say 
    1. if we currently have a core, disconnect it and clean it up
    2. create a new core of some type
    3. connect it to anything that needs connecting

- then, we need to make sure that all of the widgets are definitely using our core instance (rather than being lazy and just using `CMMCorePlus.instance()` ... in most cases i think it's being passed through, like here:
    https://github.com/pymmcore-plus/napari-micromanager/blob/603e1ddf04f6784d020264375f3625d3c98eb575/src/napari_micromanager/_gui_objects/_toolbar.py#L180-L190

    but I think i saw at least one or two cases where it wasn't being honored


------------

would you be willing to look into that?  If we had that, then we could easily add a menu item that let the user swap cores after launching the plugin.  and again, we can also discuss just having the unicore be the default one... but i think having this hygiene would be safest either way

hinderling:
Allows to swap cores via set_core(). Supports UniMMCore, automatically selects UniMMCore when loading from .cfg if #py is detected. Tests passed locally with newest version of pymmcore-plus that has support for ROI in UniMMCore (https://github.com/pymmcore-plus/pymmcore-plus/pull/550), but fail here, so I'm skipping some tests.

src/napari_micromanager/main_window.py
```diff
@@ -62,6 +80,61 @@ def __init__(
...
+        self._rebuild_toolbars(self._mmc)
+
+        # 5. Close cached dock widgets (will be recreated with new core)
+        self._close_all_dock_widgets()
+
+        # 6. Update napari console
+        if console := getattr(self.viewer.window._qt_viewer, "console", None):
+            console.push({"mmcore": self._mmc})
+
+    def _wrap_load_system_configuration(self) -> None:
```
tlambert03: :joy:  YOLO ...

let's try it out.  seems safe enough.

maybe this method should take a core instance to patch though, rather than acting as a side-effect on `self._mmc`?

tests/conftest.py
```diff
@@ -20,11 +21,36 @@
 # Prevent ipykernel debug logs from causing formatting errors in pytest
 logging.getLogger("ipykernel.inprocess.ipkernel").setLevel(logging.ERROR)
 
+# Check whether the installed UniMMCore supports getROI (needed by MDA / snap).
+# This was added on pymmcore-plus main but not yet in a pip release.
+_unicore_roi_works = True
```
tlambert03: i just released `0.17.3`... go ahead an bump the dependency in pyproject and remove these conditionals

tlambert03 (commented):
really good stuff.

I do have a couple requests for changes.   but this is a great direction.  thanks!

tlambert03:
just an additional thought here: 

there's no question that shutting down one core instance and swapping in another is a complicated operation.  So, while it's definitely still good hygiene to make sure that we're propagating the core instance all through the system, for the purposes of simply being able to use unicore (your original need here), we *can* step back and re-evaluate simpler patterns, such as an environment variable, or a "delayed" (but still single) choice of which type of core to create based on the config file.   In other words: we *wouldn't* just create a core instance at the time of MainWindow instantiation, we would do it the first time the user attempted to load a config file.  And then, after that point, the core instance would be fixed.

src/napari_micromanager/main_window.py
```diff
@@ -51,6 +69,10 @@ def __init__(
         if "MinMax" not in getattr(self.viewer.window, "dock_widgets", []):
             self.viewer.window.add_dock_widget(self.minmax, name="MinMax", area="left")
 
+        # register as the active instance
+        global _current_main_window
+        _current_main_window = self
```
tlambert03: we can probably do slightly better here.  This is a one-way update.  imagine 
1. user creates a napari viewer, and loads the plugin 
2. user creates a second napari viewer (without closing the first) and loads the plugin, this time with unicore, and ends up with a new core instance.
3. user closes the second napari viewer

you now have two problems:
1. a memory leak due to proper garbage collection of all the Qt-related resources held onto by that second window by nature of having retained a module level reference to a widget instance
2. `get_core` returns the second core that was created, even though the user is now looking back at the first napari viewer with the first core...

This stuff is tricky :joy: ... one option is to make a dynamic function that finds the current window.  That's made harder by the fact that, when you switch over to jupyter, Qt things that ALL windows are inactive and don't have focus, so you sort of have to use napari's current_viewer, but there the option to inspect qt internals is deprecated... so, this ugly thing would be the closest mess:

```python
def get_core() -> CMMCorePlus | None:
    if viewer:=napari.current_viewer()
        try:
            main_win = viewer.window._qt_window  # private access, may break
        except AttributeError:
            ...
        else:
            if our_win := main_win.findChild(napari_micromanager.MainWindow):
                return our_win.core
```

haven't really tested that... but it's an idea.  the current one will also work for simple cases, but it gives the "appearance" that it works for multiple instances... and it doesn't really

src/napari_micromanager/main_window.py
```diff
@@ -25,18 +25,47 @@
...
+    except AttributeError:
+        raise RuntimeError("Could not access napari Qt window.") from None
+    win: MainWindow | None = qt_window.findChild(MainWindow)
+    if win is None:
+        raise RuntimeError("No napari-micromanager MainWindow in this viewer.")
+    return win
+
+
+def get_core() -> CMMCorePlus:
+    """Return the CMMCorePlus instance used by the active MainWindow."""
```
tlambert03:
```suggestion
    """Return the CMMCorePlus instance used by the active MainWindow.

    Raises
    ------
    RuntimeError
        If no napari viewer or  napari-micromanager MainWindow instance can be found
    """
```

tlambert03:
thanks for working on this more... kinda looks like progress?

If you've learned anything, feel free to share.  Otherwise, please don't waste too much more of your own time on this.  This is an important, though annoying, thing that pymmcore plus needs to figure out.  It's interesting that this PR exposed it, but unless you found ways to work around it here, it's not your responsibility to fix it.

tlambert03:
I'm having one additional thought here: which is perhaps exhausting given that this *seems* to have worked (but depends on branch that is failing on pymmcore plus benchmarks)

An alternative to this "dynamically swap the cores" approach could be to only allow napari-micromanager to ever have one core: determined at launch, and never changed.  For tests, we could essentially have two invocations of `pytest`, one that tests unicore, and one that tests the regular core. 

If you are anxious to just get *something* in and useable, and if you wanted to experiment with a no-core-swapping approach that works as is with the current pymmcore-plus release, then i'd happily merge that without waiting for the upstream stuff to settle

hinderling:
thanks for keeping this in mind!  
- Merged upstream/main                            
- Removed the temporary @main CI overrides
- Added weakref in the _auto_detect_load closure + delattr sentinel on unwrap
- Bumped ruff target to py310 + fixed cascading lint
- Added myself to authors

also opened #369 to fix the test_saving_mda teardown flake that affects every CI run
let me know what you think!

tlambert03 (approved):
alright!!  Let's go! 🚀 

thank you *so* much for your hard work and patience on this one.  Classic example of a good change stress-testing existing patterns and exposing some cracks.  I'm really happy about the changes that this precipitated upstream with mmcore v12 and all the event system refactors.    So, I know it took a while, but it definitely had a broader beneficial impact 👍

PR #374, opened 2026-04-15
fix: release MainWindow and CMMCorePlus on napari window close

hinderling:
## Summary

Closing the napari window didn't release the `CMMCorePlus` singleton or the
MainWindow widget. On real hardware this leaks device handles: a subsequent
Python process that tries to load the same MM cfg fails (e.g. "No Mosaic3
devices found") until the original process is killed. Reported against an
Andor Mosaic3 but applies to any adapter that takes an exclusive handle.

## Problem

`MainWindow.__init__` creates four bound-method connections that pin `self`
and, transitively, every child widget that holds `_mmc`:

- `self.destroyed.connect(self._cleanup)` — classic self-cycle: the signal
  holds the bound method, which holds `self`, which owns the signal.
  `destroyed` therefore never fires and `_cleanup` never runs.
- `atexit.register(self._cleanup)` — same cycle via atexit.
- Three `viewer.*.events.connect(self._update_max_min)` entries — only
  disconnected inside `_cleanup`, i.e. never.

After #368 (`CMMCorePlus._instance` as `weakref.ref`) the remaining strong
references were all inside napari-micromanager, so the leak is fixable here.

A secondary leak: accessing `viewer.window._qt_viewer.console` from
`MicroManagerToolbar.__init__` triggers `napari_console._capture()`, which
walks up the call stack to the first non-napari frame and pushes its whole
namespace — including the `self` local — into the IPython user_ns. That
reference persists for the process lifetime.

## Solution

**`main_window.py`**
- Register a weakref-based closure (`_weak_cleanup`) instead of the bound
  method. Neither `self.destroyed`, `atexit`, nor the new proactive hook
  pin `self` anymore.
- Connect `viewer.window._qt_window.destroyed` as the primary cleanup trigger.
  It fires while the viewer is still alive, which is the only time signal
  disconnection actually works. `self.destroyed` + `atexit` remain as
  fallbacks. If napari ever drops `_qt_window`, we `warn()` instead of
  silently regressing.
- `_cleanup` is now idempotent, and `self._connections.clear()` breaks the
  remaining `list → tuple → bound method → self` cycle after disconnects.
- Wrap `self._core_link.cleanup()` in `contextlib.suppress(Exception)` so a
  misbehaving device can't abort the rest of teardown.

**`_gui_objects/_toolbar.py`**
- Add `NAPARI_EMBED = True` at module level — napari-console's documented
  opt-out sentinel for its frame-capture behavior.

## Risk: unresponsive devices on teardown

Unlike today, `_cleanup()` now actually runs on window close. That means
`_core_link.cleanup()` → `mmc.stopSequenceAcquisition()` will now fire
against the camera adapter while the viewer is closing. If live-mode is
running on a wedged device, that C-level call can raise or block — the
`suppress(Exception)` handles the raise case, but a hang would appear as
a zombie process (same symptom as the bug this PR fixes, just for a
different reason). Net effect is still strictly better: before this PR
the process *always* lingered; after this PR it only lingers on a wedged
device.

The long-term mitigation belongs at the pymmcore-plus boundary, not here —
see pymmcore-plus/pymmcore-plus#614, which proposes per-device timeouts on
the wait calls that sit underneath `stopSequenceAcquisition`. If that
lands, the stop path can fail-fast instead of hanging. Not a dependency
for this PR.

## Test plan

- [x] New regression test: `tests/test_teardown.py::test_main_window_released_after_viewer_close`.
  Creates a viewer + MainWindow, closes, pumps the Qt event loop, asserts
  that a weakref to MainWindow resolves to `None`. Fails cleanly with a
  `gc.get_referrers` dump of the 5 surviving bound-method referrers before
  the fix; passes after.
- [x] Full suite: 61 passed, 1 skipped, 1 pre-existing failure on main
  (`test_set_core.py::test_auto_detect_standard_cfg_swaps_to_mmcore`, fails
  on bare upstream too — unrelated to this PR).
- [x] On-hardware verification (Mosaic3 cfg, Shell 1 / Shell 3 sequence from
  the original report) — confirms the exclusive handle is actually released.
  Not runnable in CI.

## Notes on scope

- This fix only touches napari-micromanager cycles. The lingering
  `setExposure` bound method and `mmcore` key in the IPython user_ns live
  in pymmcore-widgets and napari-console respectively and don't block the
  main leak from being fixed here.
- Reusing `mmc` from an external notebook across an open/close cycle still
  works: `_cleanup` only disconnects our own listeners; it doesn't call
  `unloadAllDevices()` or otherwise mutate core state.

### DIFF
--- a/src/napari_micromanager/_gui_objects/_toolbar.py
+++ b/src/napari_micromanager/_gui_objects/_toolbar.py
@@ -80,7 +80,19 @@ class MicroManagerToolbar(QMainWindow):
     ) -> None:
         super().__init__()
 
-        self._mmc = mmcore or CMMCorePlus.instance()
+        # If no core is passed in, the plugin creates its own fresh core and
+        # takes responsibility for its lifecycle (cancel MDAs, unload devices
+        # on close). Deliberately avoid `CMMCorePlus.instance()` — that would
+        # silently adopt a singleton the plugin didn't create, making
+        # ownership ambiguous. To have the plugin drive an existing core
+        # without taking ownership, pass it as `mmcore=`; the plugin will
+        # leave its lifecycle to the caller.
+        if mmcore is None:
+            self._mmc = CMMCorePlus()
+            self._owns_core = True
+        else:
+            self._mmc = mmcore
+            self._owns_core = False
         self.viewer: napari.viewer.Viewer = getattr(viewer, "__wrapped__", viewer)
 
         # add variables to the napari console
--- a/src/napari_micromanager/_gui_objects/_toolbar.py
+++ b/src/napari_micromanager/_gui_objects/_toolbar.py
@@ -48,6 +48,11 @@ from napari_micromanager._gui_objects._stages_widget import MMStagesWidget
 if TYPE_CHECKING:
     import napari.viewer
 
+# Opt out of napari-console's `_capture()` — without this, creating the console
+# below would push this module's frame (including `self`) into the IPython
+# user_ns, pinning MainWindow for the process lifetime.
+NAPARI_EMBED = True
+
 TOOL_SIZE = 35
 
 
--- a/src/napari_micromanager/_gui_objects/_toolbar.py
+++ b/src/napari_micromanager/_gui_objects/_toolbar.py
@@ -48,9 +48,13 @@ from napari_micromanager._gui_objects._stages_widget import MMStagesWidget
 if TYPE_CHECKING:
     import napari.viewer
 
-# Opt out of napari-console's `_capture()` — without this, creating the console
-# below would push this module's frame (including `self`) into the IPython
-# user_ns, pinning MainWindow for the process lifetime.
+# Opt out of napari-console's `_capture()`. This is an un/mis-documented API
+# in `napari_console.qt_console`: the napari-console README says to set an
+# environment variable `NAPARI_EMBED=1`, but the actual check in
+# `QtConsole._capture` looks for `NAPARI_EMBED` in the calling frame's module
+# globals. Without this opt-out, creating the console below would push this
+# module's frame (including the `self` local of `MicroManagerToolbar.__init__`)
+# into the IPython user_ns, pinning MainWindow for the process lifetime.
 NAPARI_EMBED = True
 
 TOOL_SIZE = 35
--- a/src/napari_micromanager/main_window.py
+++ b/src/napari_micromanager/main_window.py
@@ -75,9 +75,10 @@ class MainWindow(MicroManagerToolbar):
         self,
         viewer: napari.viewer.Viewer,
         config: str | Path | None = None,
+        mmcore: CMMCorePlus | None = None,
     ) -> None:
-        super().__init__(viewer)
-        self.set_core(self._mmc)
+        super().__init__(viewer, mmcore=mmcore)
+        self.set_core(self._mmc, owns=self._owns_core)
 
         # some remaining connections related to widgets ... TODO: unify with superclass
         self._connections: list[tuple[PSignalInstance, Callable]] = [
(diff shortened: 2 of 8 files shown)


src/napari_micromanager/_gui_objects/_toolbar.py
```diff
@@ -48,6 +48,11 @@
 if TYPE_CHECKING:
     import napari.viewer
 
+# Opt out of napari-console's `_capture()` — without this, creating the console
+# below would push this module's frame (including `self`) into the IPython
+# user_ns, pinning MainWindow for the process lifetime.
+NAPARI_EMBED = True
```
tlambert03, line 58: is this actually doing anything?  I'm unfamiliar with it, googled it and the napari-console readme says

> If you want to disable this behavior (for example, because you are embedding napari and the console within some larger application), add `NAPARI_EMBED=1` to your environment variables before instantiating the console.

but this isn't doing that right?

src/napari_micromanager/main_window.py
```diff
@@ -91,9 +92,32 @@ def __init__(
...
+                "napari viewer has no `_qt_window`; eager cleanup on window "
+                "close is disabled and device handles may leak until process "
+                "exit.",
+                stacklevel=2,
+            )
+        else:
+            qt_window.destroyed.connect(_weak_cleanup)
+
+        self.destroyed.connect(_weak_cleanup)
+        atexit.register(_weak_cleanup)
```
tlambert03, line 122: having 3 belt & suspenders cleanups feels excessive.  Do you happen to know which, if any, are *actually* helping?

src/napari_micromanager/main_window.py
```diff
@@ -91,9 +92,32 @@ def __init__(
         if "MinMax" not in getattr(self.viewer.window, "dock_widgets", []):
             self.viewer.window.add_dock_widget(self.minmax, name="MinMax", area="left")
 
-        # queue cleanup
-        self.destroyed.connect(self._cleanup)
```
tlambert03, line 95: btw, general comment: 

this pattern of `self.destroyed.connect(self.method)` was probably always broken and never doing what is looks like it's doing.  Using `atexit.register` or `weakref.finalize` is preferable (and even then, should probably be used sparingly).  So, you can probably get rid of the `self.destroyed.connect` call altogether (probably should).

tlambert03 (approved):
some side questions and comments.  but none of them blocking.  Your actual tests with real hardware are the key here so if this works for you, feel free to merge whenever you're happy with the solution

src/napari_micromanager/main_window.py
hinderling: Dropped `self.destroyed` in c89fb36 per your other comment. Down to `qt_window.destroyed` (primary, only one that actually fires on close — instrumented) + `atexit` (process-exit fallback).

src/napari_micromanager/main_window.py
hinderling: Dropped `self.destroyed` in c89fb36. Down to `qt_window.destroyed` (primary, only one that actually fires on close — instrumented) + `atexit` (process-exit fallback).

src/napari_micromanager/_gui_objects/_toolbar.py
hinderling: README seems misleading. The actual check is `"NAPARI_EMBED" not in c.frame.f_globals` ([qt_console.py:182](https://github.com/napari/napari-console/blob/main/napari_console/qt_console.py#L182)), not `os.environ`. A module-level `NAPARI_EMBED = True` is what it looks for. Removing it makes the teardown test fail because napari-console's `_capture()` walks up to the first non-napari frame (our `_toolbar.py` `__init__`), pushes that frame's locals + globals into the IPython user_ns, and `self` ends up pinned there for the process lifetime.

src/napari_micromanager/main_window.py
tlambert03: ok, yeah: readme is wrong.  this is a truly hard to discover API, can you add a comment above napari_embed explaining that it's an un/mis-documented API in `napari_console.qt_console`

hinderling: Expanded the comment in e8e9f02 to spell out the README-vs-code discrepancy and where the actual check lives.

hinderling:
Testing on real hardware, there were abandoned python threads that kept devices busy even after closing the window. Now explicitly calling `_mmc.unloadAllDevices()` and `_mmc.mda.cancel()`. With the new functionality of being able to set cores from e.g. notebooks (https://github.com/pymmcore-plus/napari-micromanager/pull/365), in this case we likely only want to use napari-micromanager as a temporary viewer, and not unload devices if the window is closed, and also not cancel any MDA sequences that were started programatically. So there is a notion of "ownership". The plugin now has bool `_owns_core` that is true if the core was instantiated via the toolbar, `set_core` calls now has arg `owns=False`; if we pass a core to the plugin from the outside, the outside-script "retains ownership" of the core, closing the napari plugin in this case does not shut down the microscope. Another change that follows from this is that we shouldn't call `CMMCorePlus.instance()`, as this leads to unclear ownership. `__init__` now takes `mmcore: CMMCorePlus | None` to avoid `.instance()` calls. I can see that this is biased to our usecase, but think it's in line with the discussions we had in https://github.com/pymmcore-plus/napari-micromanager/pull/365 regarding core hygiene; one thing we decided to remove there and that I decided to still add here is passing mmcore to init. In practice I think it's cleaner (avoids an extra `set_core` line when starting the plugin interactively, and avoids creation of throwaway core).

With these changes, everything shuts down cleanly now on our hardware, no more hanging devices. But happy for critique!

PR #376, opened 2026-05-18
fix: tolerate SystemError from disconnect() in _cleanup teardown

hinderling:
## Summary

The three `_cleanup()` teardown loops — in `MainWindow`, `CoreViewerLink`, and `_NapariMDAHandler` — disconnect their core signal/slot pairs inside `contextlib.suppress(TypeError, RuntimeError)`, tolerating an already-disconnected slot.

PySide6's `SignalInstance.disconnect()` can raise **`SystemError`** for that no-op case:

```
SystemError: <method 'disconnect' of 'PySide6.QtCore.SignalInstance' objects>
             returned a result with an exception set
```

That exception type is not in the `suppress` tuple, so it escapes and propagates out of `_cleanup()`.

## Not a recent regression

This is **long-standing PySide6 behaviour**, not new in any particular version. It was reported against **PySide6 6.7.1** in [pytest-qt#558](https://github.com/pytest-dev/pytest-qt/issues/558) back in 2024, and `pytest-qt` handles it exactly the same way — `SystemError` is included in its `_silent_disconnect` catch.

It surfaces in CI now only because the `windows-latest / py3.13 / PySide6 / highest` matrix cell resolves to a PySide6 version that exercises this path — `tests/test_layer_scale.py` fails at teardown with the `SystemError`. It's unrelated to any in-flight feature work: `main` fails the same way on its next CI run.

## Fix

Add `SystemError` to the suppressed exceptions in all three cleanup loops. It's another form of the same already-disconnected condition the existing `TypeError` / `RuntimeError` entries already cover.

## Notes

- 3-line change, no behaviour change on PyQt or on PySide6 builds that don't raise `SystemError` here (it's a literal no-op there — `suppress` only catches what's actually raised).
- Regression coverage is already implicit: the `PySide6 / highest` CI cell exercises these teardown paths via the existing suite — removing `SystemError` would fail it again there.

## Test plan

- [x] CI green on `windows-latest / py3.13 / PySide6 / highest` (previously failing on `test_layer_scale.py`)
- [x] No regression on the PyQt5 / PyQt6 / PySide6 (lowest-direct & highest) matrix cells

## References

- [pytest-qt#558 — *SystemError on disconnect with pyside 6.7.1*](https://github.com/pytest-dev/pytest-qt/issues/558) (same error; filed 2024)
- [pytest-qt#552 — *"RuntimeError: Failed to disconnect" with PySide6==6.7.0*](https://github.com/pytest-dev/pytest-qt/issues/552) (related, earlier exception form)

### DIFF
--- a/src/napari_micromanager/_core_link.py
+++ b/src/napari_micromanager/_core_link.py
@@ -59,7 +59,7 @@ class CoreViewerLink(QObject):
                 self._mmc.stopSequenceAcquisition()
 
         for signal, slot in self._connections:
-            with contextlib.suppress(TypeError, RuntimeError, SystemError):
+            with contextlib.suppress(Exception):
                 signal.disconnect(slot)
         # Clean up temporary files we opened.
         self._mda_handler._cleanup()
--- a/src/napari_micromanager/_core_link.py
+++ b/src/napari_micromanager/_core_link.py
@@ -59,7 +59,7 @@ class CoreViewerLink(QObject):
                 self._mmc.stopSequenceAcquisition()
 
         for signal, slot in self._connections:
-            with contextlib.suppress(TypeError, RuntimeError):
+            with contextlib.suppress(TypeError, RuntimeError, SystemError):
                 signal.disconnect(slot)
         # Clean up temporary files we opened.
         self._mda_handler._cleanup()
--- a/src/napari_micromanager/_mda_handler.py
+++ b/src/napari_micromanager/_mda_handler.py
@@ -83,7 +83,7 @@ class _NapariMDAHandler:
     def _cleanup(self) -> None:
         self._mda_running = False  # stops the worker thread loop
         for signal, slot in self._connections:
-            with contextlib.suppress(TypeError, RuntimeError):
+            with contextlib.suppress(TypeError, RuntimeError, SystemError):
                 signal.disconnect(slot)
         # Clean up temporary files we opened.
         for z, v in self._tmp_arrays.values():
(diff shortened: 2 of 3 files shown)


src/napari_micromanager/_core_link.py
```diff
@@ -59,7 +59,7 @@ def cleanup(self) -> None:
                 self._mmc.stopSequenceAcquisition()
 
         for signal, slot in self._connections:
-            with contextlib.suppress(TypeError, RuntimeError):
+            with contextlib.suppress(TypeError, RuntimeError, SystemError):
```
tlambert03:
```suggestion
            with contextlib.suppress(Exception):
```

i can't think of an exception that we really do care about here.  let's just be broader

src/napari_micromanager/_mda_handler.py
```diff
@@ -83,7 +83,7 @@ def __init__(self, mmcore: CMMCorePlus, viewer: napari.viewer.Viewer) -> None:
     def _cleanup(self) -> None:
         self._mda_running = False  # stops the worker thread loop
         for signal, slot in self._connections:
-            with contextlib.suppress(TypeError, RuntimeError):
+            with contextlib.suppress(TypeError, RuntimeError, SystemError):
```
tlambert03:
```suggestion
            with contextlib.suppress(Exception):
```

src/napari_micromanager/main_window.py
```diff
@@ -183,7 +183,7 @@ def _unwrap_load_system_configuration(self, core: CMMCorePlus) -> None:
     def _cleanup(self) -> None:
         self._unwrap_load_system_configuration(self._mmc)
         for signal, slot in self._connections:
-            with contextlib.suppress(TypeError, RuntimeError):
+            with contextlib.suppress(TypeError, RuntimeError, SystemError):
```
tlambert03:
```suggestion
            with contextlib.suppress(Exception):
```

